PR #6, opened 2017-09-06
JAV-320 configurable recovery policy

zhengyangyong:


saga-core/src/main/java/io/servicecomb/saga/core/application/SagaCoordinator.java
```diff
@@ -74,23 +84,29 @@ public SagaCoordinator(
...
     this.persistentStore = persistentStore;
     this.requestInterpreter = new JsonRequestInterpreter(fromJsonFormat, new GraphCycleDetectorImpl<>());
+    this.fromJsonFormat = fromJsonFormat;
     this.toJsonFormat = toJsonFormat;
     this.executorService = executorService;
   }
 
   @Segment(name = "runSagaCoordinator", category = "application", library = "kamon")
-  public void run(String requestJson) {
+  public void run(String requestJson) throws IOException {
```
seanyinx: we shouldn't throw checked exception out

saga-core/src/main/java/io/servicecomb/saga/core/application/SagaCoordinator.java
```diff
@@ -105,12 +121,14 @@ public void reanimate() {
       Saga saga = new Saga(
           eventStore,
           executorService,
-          recoveryPolicy,
+          recoveryPolicy.get(event.sagaId),
```
seanyinx: this will not work, since this method is called on restart

saga-core/src/test/java/io/servicecomb/saga/core/SagaCoordinatorTest.java
```diff
@@ -135,8 +168,20 @@ public void runSagaWithEventStore() {
 
   @Test
   public void processRequestsInParallel() {
-    CompletableFuture.runAsync(() -> coordinator.run(requestJson));
-    CompletableFuture.runAsync(() -> coordinator.run(anotherRequestJson));
+    CompletableFuture.runAsync(() -> {
```
seanyinx: coordinator is not supposed to throw checked exception

saga-format/src/main/java/io/servicecomb/saga/format/JsonSagaDefinition.java
```diff
@@ -0,0 +1,42 @@
...
+package io.servicecomb.saga.format;
+
+import com.fasterxml.jackson.annotation.JsonRawValue;
+import com.fasterxml.jackson.databind.JsonNode;
+import io.servicecomb.saga.core.SagaDefinition;
+
+public class JsonSagaDefinition implements SagaDefinition {
+
+  private String policy;
+  private Object requests;
```
seanyinx: why not deserialize the requests too?

saga-core/src/main/java/io/servicecomb/saga/core/application/SagaCoordinator.java
zhengyangyong: Done

saga-core/src/test/java/io/servicecomb/saga/core/SagaCoordinatorTest.java
zhengyangyong: Done

saga-format/src/main/java/io/servicecomb/saga/format/JsonSagaDefinition.java
zhengyangyong: 原来构建Task和Event都是使用的String,暂时不修改了

saga-core/src/main/java/io/servicecomb/saga/core/application/SagaCoordinator.java
zhengyangyong: Done

PR #7, opened 2017-09-08
JAV-201 added service discovery transport with service center

seanyinx:


pom.xml
```diff
@@ -87,6 +91,16 @@
         <version>0.0.1-SNAPSHOT</version>
       </dependency>
       <dependency>
+        <groupId>io.servicecomb.saga.transports</groupId>
+        <artifactId>transport-httpclient-spring</artifactId>
```
WillemJiang: Does this module have something to do with spring?

seanyinx: yes, to expose underlying transport implementation as a spring bean

PR #23, opened 2017-09-21
updated saga.version

zhang2014:
#21 @seanyinx


seanyinx:
@zhang2014 very elegant solution to refer saga.version to project.version

Issue #17, opened 2017-09-18
scenario on db failure


seanyinx:
what to do with db failure in middle of transactions?
compensate or retry?

MabinGo:
Maybe  it's better to do retrying all the time until the database is recovered, cause all the event is store in db.what's your opinion?

seanyinx:
there seems no better way than keep retrying till database is recovered, if some transactions in the request have already started.

if none of the transactions is started, we can safely reject the request.

PR #24, opened 2017-09-23
JAV-328 handle scenario on db failure. 

MabinGo:
reference issue #17 
Signed-off-by: MabinGo <bin.ma@huawei.com>

### DIFF
--- a/saga-core/src/main/java/io/servicecomb/saga/core/application/SagaExecutionComponent.java
+++ b/saga-core/src/main/java/io/servicecomb/saga/core/application/SagaExecutionComponent.java
@@ -142,11 +142,11 @@ public class SagaExecutionComponent {
 
   static class RetrySagaLog implements PersistentStore {
 
-    private final PersistentStore retryPersistentStore;
+    private final PersistentStore persistentStore1;
     private final int retryDelay;
 
     RetrySagaLog(PersistentStore persistentStore, int retryDelay) {
-      this.retryPersistentStore = persistentStore;
+      this.persistentStore1 = persistentStore;
       this.retryDelay = retryDelay;
     }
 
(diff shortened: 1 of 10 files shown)


saga-core/src/main/java/io/servicecomb/saga/core/application/SagaExecutionComponent.java
```diff
@@ -122,17 +125,57 @@ public void reanimate() {
...
-      put(SAGA_END_TASK, new SagaEndTask(sagaId, compositeSagaLog));
+      SagaLog retrySagaLog = compositeSagaLog(sagaLog, SagaExecutionComponent.this.retrySagaLog);
+      put(SAGA_REQUEST_TASK, new RequestProcessTask(sagaId, retrySagaLog, fallbackPolicy));
+      put(SAGA_END_TASK, new SagaEndTask(sagaId, retrySagaLog));
     }};
   }
+
+  static class RetrySagaLog implements PersistentStore {
+
+    private final PersistentStore retryPersistentStore;
```
seanyinx: this naming seems not fit to the actual persistent store

saga-core/src/main/java/io/servicecomb/saga/core/application/SagaExecutionComponent.java
```diff
@@ -122,17 +125,57 @@ public void reanimate() {
...
+      do {
+        try {
+          retryPersistentStore.offer(sagaEvent);
+          success = true;
+        } catch (Exception e) {
+          e.printStackTrace();
+          try {
+            sleep(retryDelay);
+          } catch (InterruptedException e1) {
+            e1.printStackTrace();
```
seanyinx: when this exception is thrown, normally it means this application is shutting down, because we didn't manually interrupt threads anywhere in code.

when saga is shutting down, we have to exit the while loop, or saga won't be able to stop. you may refer to FallbackPolicy on how to handle this exception.

saga-core/src/test/java/io/servicecomb/saga/core/BackwardRecoveryTest.java
```diff
@@ -0,0 +1,28 @@
...
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+
+import org.junit.Test;
+
+public class BackwardRecoveryTest {
+
+  private final SagaTask sagaTask = mock(SagaTask.class);
+  private final SagaRequest sagaRequest = mock(SagaRequest.class);
+  private final
```
seanyinx: the line break seems to be wrong

saga-core/src/test/java/io/servicecomb/saga/core/SagaStartTaskTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.Test;
+
+public class SagaStartTaskTest {
+  private final SagaRequest request = mock(SagaRequest.class);
+  private final SagaLog sagaLog = mock(SagaLog.class);
+
+  private final SagaStartTask sagaStartTask = new SagaStartTask("0", null, sagaLog);
+
+  @Test
+  public void commit() {
```
seanyinx: it's better to provide more meaningful name to reveal your intention of this test, eg. blowsUpWhenEventIsNotPersisted, so that it's easier for readers to understand the purpose of this test. you may refer to other tests.

saga-core/src/test/java/io/servicecomb/saga/core/application/RetrySagaLogTest.java
```diff
@@ -0,0 +1,45 @@
...
+  private List<String> list = new ArrayList<>();
+
+  @Test
+  public void offer() throws InterruptedException {
+
+    ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
+    executor.execute(() -> {
+      doThrow(RuntimeException.class).when(retryPersistentStore).offer(dummyEvent);
+      retrySagaLog.offer(dummyEvent);
+      list.add("failure");
```
seanyinx: this line can never be reached, because retry saga log will never return. that means the logic below to check is list is empty will always be false

PR #27, opened 2017-10-09
Jav 370 forward recovery delay

zhengyangyong:
最大变更是修改了RecoveryPolicy接口的apply方法由void返回为retryCount，用于单元测试；
虽然有别的方式，但是觉得不好看，retryCount下一个阶段会有用，例如需要扩展ForwardRecovery不能无限重试，可以设置重试次数，然后做检查。

### DIFF
--- a/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
+++ b/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
@@ -25,30 +25,26 @@ public class ForwardRecovery implements RecoveryPolicy {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   @Override
-  public int apply(SagaTask task, SagaRequest request) {
-    boolean success = false;
-    int retryCount = 0;
-    do {
-      try {
-        task.commit(request);
-        success = true;
-      } catch (SagaStartFailedException e) {
-        throw e;
-      } catch (Exception e) {
-        log.error("Applying {} policy due to failure in transaction {} of service {}",
-            description(),
-            request.transaction(),
-            request.serviceName(),
-            e
-        );
+  public void apply(SagaTask task, SagaRequest request) {
+    try {
+      boolean success = false;
+      do {
         try {
-          Thread.sleep((int) (request.failRetryDelaySeconds() * 1000));
-        } catch (InterruptedException ie) {
-          break;
+          task.commit(request);
+          success = true;
+        } catch (SagaStartFailedException e) {
+          throw e;
+        } catch (Exception e) {
+          log.error("Applying {} policy due to failure in transaction {} of service {}",
+              description(),
+              request.transaction(),
+              request.serviceName(),
+              e
+          );
+          Thread.sleep(request.failRetryDelayMilliseconds());
         }
-        retryCount++;
-      }
-    } while (!success);
-    return retryCount;
+      } while (!success);
+    } catch (InterruptedException ignored) {
+    }
   }
 }
--- a/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
+++ b/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
@@ -25,9 +25,9 @@ public class ForwardRecovery implements RecoveryPolicy {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   @Override
-  public void apply(SagaTask task, SagaRequest request) {
+  public int apply(SagaTask task, SagaRequest request) {
     boolean success = false;
-
+    int retryCount = 0;
     do {
       try {
         task.commit(request);
@@ -42,11 +42,13 @@ public class ForwardRecovery implements RecoveryPolicy {
             e
         );
         try {
-          Thread.sleep(request.failRetryDelaySeconds() * 1000);
+          Thread.sleep((int) (request.failRetryDelaySeconds() * 1000));
         } catch (InterruptedException ie) {
-          return;
+          break;
         }
+        retryCount++;
       }
     } while (!success);
+    return retryCount;
   }
 }
--- a/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
+++ b/saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
@@ -41,6 +41,11 @@ public class ForwardRecovery implements RecoveryPolicy {
             request.serviceName(),
             e
         );
+        try {
+          Thread.sleep(request.failRetryDelaySeconds() * 1000);
+        } catch (InterruptedException ie) {
+          return;
+        }
       }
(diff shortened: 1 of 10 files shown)


saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -41,7 +41,14 @@ public void apply(SagaTask task, SagaRequest request) {
             request.serviceName(),
             e
         );
+        try {
+          Thread.sleep((int) (request.failRetryDelaySeconds() * 1000));
+        } catch (InterruptedException ie) {
+          break;
```
seanyinx: why not catch interruption outside while loop, in order to avoid break again?

saga-core/src/main/java/io/servicecomb/saga/core/SagaRequestImpl.java
```diff
@@ -30,6 +30,7 @@
   private final Compensation compensation;
   private final String[] parents;
   private final Fallback fallback;
+  private final double failRetryDelaySeconds;
```
seanyinx: why double?

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -41,7 +41,14 @@ public void apply(SagaTask task, SagaRequest request) {
             request.serviceName(),
             e
         );
+        try {
+          Thread.sleep((int) (request.failRetryDelaySeconds() * 1000));
```
seanyinx: preferable to configure delay in milliseconds

saga-core/src/main/java/io/servicecomb/saga/core/SagaRequestImpl.java
```diff
@@ -55,6 +58,17 @@ public SagaRequestImpl(
       String type,
       Transaction transaction,
       Compensation compensation,
+      Fallback fallback,
+      String[] parents) {
+    this(id, serviceName, type, transaction, compensation, fallback, parents, 0);
```
seanyinx: there shall be a default delay greater than zero, preferably a few seconds to avoid performance impact

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -25,9 +25,9 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   @Override
-  public void apply(SagaTask task, SagaRequest request) {
+  public int apply(SagaTask task, SagaRequest request) {
     boolean success = false;
-
+    int retryCount = 0;
```
seanyinx: if retry count is only for testing, it's a bit overkill to change signature for this purpose.

you can verify how many times the task mock is invoked instead.

zhengyangyong: Done

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
zhengyangyong: JAV上的描述是秒，已经更改为毫秒

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
zhengyangyong: Done

saga-core/src/main/java/io/servicecomb/saga/core/SagaRequestImpl.java
zhengyangyong: Done

saga-core/src/main/java/io/servicecomb/saga/core/SagaRequestImpl.java
zhengyangyong: Done，默认为50毫秒

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -26,22 +26,25 @@
...
+              description(),
+              request.transaction(),
+              request.serviceName(),
+              e
+          );
+          Thread.sleep(request.failRetryDelayMilliseconds());
+        }
+      } while (!success);
+    } catch (InterruptedException ignored) {
+    }
```
seanyinx: preferably log a warning in catch block to make troubleshooting easier.

saga-format/src/main/java/io/servicecomb/saga/format/JsonRestSagaRequest.java
```diff
@@ -38,16 +38,19 @@ public JsonRestSagaRequest(
...
       @JsonProperty("compensation") JacksonRestCompensation compensation,
       @JsonProperty("fallback") JacksonFallback fallback,
-      @JsonProperty("parents") String[] parents) {
+      @JsonProperty("parents") String[] parents,
+      @JsonProperty("failRetryDelayMilliseconds") int failRetryDelayMilliseconds) {
 
-    super(id, serviceName, type, transaction, compensation, fallback == null? NOP_TRANSPORT_AWARE_FALLBACK : fallback, parents);
+    super(id, serviceName, type, transaction, compensation,
+        fallback == null ? NOP_TRANSPORT_AWARE_FALLBACK : fallback,
+        parents, failRetryDelayMilliseconds);
```
seanyinx: jackson json writes 0 as default value for int. have to check here for 0 and negative (in case of human error) and pass a constant we defined as default here.

saga-core/src/main/java/io/servicecomb/saga/core/ForwardRecovery.java
zhengyangyong: Done

saga-format/src/main/java/io/servicecomb/saga/format/JsonRestSagaRequest.java
zhengyangyong: Done

saga-core/src/main/java/io/servicecomb/saga/core/SagaRequestImpl.java
```diff
@@ -38,14 +39,16 @@ public SagaRequestImpl(
...
+      String[] parents,
+      int failRetryDelayMilliseconds) {
 
     this.id = id;
     this.serviceName = serviceName;
     this.type = type;
     this.transaction = transaction;
     this.compensation = compensation;
     this.fallback = fallback;
+    this.failRetryDelayMilliseconds = failRetryDelayMilliseconds == 0 ? 50 : failRetryDelayMilliseconds;
```
seanyinx: if delay is negative, thread.sleep will throw exception. we have to either throw now or use default delay too.

zhengyangyong: OK

PR #28, opened 2017-10-11
Jav 315 response out

seanyinx:


saga-core/src/main/java/io/servicecomb/saga/core/CompositeSagaResponse.java
```diff
@@ -0,0 +1,27 @@
+package io.servicecomb.saga.core;
```
WillemJiang: Please add the license header for the new added file.

saga-core/src/main/java/io/servicecomb/saga/core/PersistentLog.java
```diff
@@ -0,0 +1,5 @@
+package io.servicecomb.saga.core;
```
WillemJiang: license header.

saga-core/src/main/java/io/servicecomb/saga/core/SagaContext.java
```diff
@@ -0,0 +1,21 @@
+package io.servicecomb.saga.core;
```
WillemJiang: license header.

saga-core/src/main/java/io/servicecomb/saga/core/CompositeSagaResponse.java
seanyinx: done

saga-core/src/main/java/io/servicecomb/saga/core/PersistentLog.java
seanyinx: done

saga-core/src/main/java/io/servicecomb/saga/core/SagaContext.java
seanyinx: done

seanyinx: done

Issue #20, opened 2017-09-20
latency & fault tolerance in transactions/compensations


seanyinx:
*Given*
* saga is up and running

*When*
* a transaction request involves service a
* service a is unreachable or very slow

*Then*
* the transaction times out with pre configured timeout interval

Issue #19, opened 2017-09-20
add timeout for httpclient


seanyinx:
it seems rest request with http client is not timing out

seanyinx:
this should be handled by https://github.com/ServiceComb/ServiceComb-Saga/issues/20

PR #53, opened 2017-10-28
added timeout settings for rest transport

eric-lee-ltk:
solved issue #20 
Signed-off-by: eric-lee-ltk <lidagang1@huawei.com>

### DIFF
--- a/transports/transport-httpclient/pom.xml
+++ b/transports/transport-httpclient/pom.xml
@@ -47,17 +47,14 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock-standalone</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.github.seanyinx</groupId>
       <artifactId>unit-scaffolding</artifactId>
-      <scope>test</scope>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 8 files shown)


transports/transport-httpclient/pom.xml
```diff
@@ -47,14 +47,17 @@
...
     </dependency>
     <dependency>
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock-standalone</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.github.seanyinx</groupId>
       <artifactId>unit-scaffolding</artifactId>
+      <scope>test</scope>
```
seanyinx: scope is inherited from parent and not necessary if defined in parent pom

transports/transport-httpclient/src/main/java/io/servicecomb/saga/transports/httpclient/HttpClientTransport.java
```diff
@@ -28,18 +24,26 @@
...
+import io.servicecomb.saga.core.SuccessfulSagaResponse;
+import io.servicecomb.saga.core.TransportFailedException;
+import io.servicecomb.saga.transports.RestTransport;
+import kamon.annotation.EnableKamon;
+import kamon.annotation.Segment;
+
 @EnableKamon
 public class HttpClientTransport implements RestTransport {
 
+  private int requestTimeout;
```
seanyinx: this field can be final. making it apparent to user that it's immutable

seanyinx:
@eric-lee-ltk this actually solves #19 
we need to integrate hystrix for #20

Issue #59, opened 2017-11-01
gracefully shutdown of actor system


seanyinx:
currently actor system is not properly shut down

PR #78, opened 2017-12-18
gracefully shutdown of actor system

zhang2014:
#59 @seanyinx


saga-core/src/main/java/io/servicecomb/saga/core/dag/GraphBasedSagaFactory.java
```diff
@@ -34,6 +35,7 @@
 import io.servicecomb.saga.infrastructure.ContextAwareEventStore;
 
 public class GraphBasedSagaFactory implements SagaFactory {
+  private final AtomicBoolean isRunning = new AtomicBoolean(true);
```
seanyinx: is a `volatile boolean` enough?

seanyinx:
fix https://github.com/ServiceComb/ServiceComb-Saga/issues/59

PR #84, opened 2017-12-20
add license for SagaShutdownListener

zhang2014:
#83 @seanyinx


seanyinx:
fix https://github.com/ServiceComb/ServiceComb-Saga/issues/83

PR #87, opened 2017-12-25
[SCB-96] tx interception

seanyinx:


omega/omega-transport/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
WillemJiang: We need to add the license header here.

PR #95, opened 2018-01-02
[SCB-149] service aware callback

seanyinx:

---


alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/CompositeOmegaCallback.java
```diff
@@ -0,0 +1,44 @@
...
+  public CompositeOmegaCallback(Map<String, Map<String, OmegaCallback>> callbacks) {
+    this.callbacks = callbacks;
+  }
+
+  @Override
+  public void compensate(TxEvent event) {
+    Map<String, OmegaCallback> serviceCallbacks = callbacks.get(event.serviceName());
+
+    if (serviceCallbacks.isEmpty()) {
+      throw new AlphaException("No such omega callback found for service " + event.serviceName());
```
WillemJiang: Here we just know about the service name, I think we need to put the event here.

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/CompositeOmegaCallback.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    OmegaCallback omegaCallback = serviceCallbacks.get(event.instanceId());
+    if (omegaCallback == null) {
+      serviceCallbacks.values().iterator().next().compensate(event);
```
WillemJiang: It looks like a default compensate call back here.  How about there are different instance of service which need to be compensated?

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/RetryOmegaCallback.java
```diff
@@ -0,0 +1,71 @@
...
+  private final OmegaCallback underlying;
+  private final int delay;
+
+  public RetryOmegaCallback(OmegaCallback underlying, int delay) {
+    this.underlying = underlying;
+    this.delay = delay;
+  }
+
+  @Override
+  public void compensate(TxEvent event) {
```
WillemJiang: The calling thread is blocked, we may need to think for another way to do it.

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/CompositeOmegaCallback.java
seanyinx: what event?

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/CompositeOmegaCallback.java
seanyinx: there are multiple instances of the same service stored in the map, we just choose the first instance, if the matching instance is unreachable and removed from the map

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/RetryOmegaCallback.java
seanyinx: the compensation is done in a separate thread from the event handling thread.

alpha/alpha-core/src/main/java/io/servicecomb/saga/alpha/core/CompositeOmegaCallback.java
WillemJiang: How about add the event to the exception, or log the event.

PR #100, opened 2018-01-03
[SCB-100] async support

seanyinx:

---

### DIFF
--- a/omega/omega-spring-tx/src/main/java/org/apache/servicecomb/saga/omega/transaction/spring/ExecutorFieldCallback.java
+++ b/omega/omega-spring-tx/src/main/java/org/apache/servicecomb/saga/omega/transaction/spring/ExecutorFieldCallback.java
@@ -1,0 +1,163 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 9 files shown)


omega/omega-spring-tx/src/main/java/org/apache/servicecomb/saga/omega/transaction/spring/ExecutorFieldCallback.java
```diff
@@ -0,0 +1,167 @@
...
+
+    field.set(bean, ExecutorProxy.newInstance(field.get(bean), field.getType(), omegaContext));
+  }
+
+  private static class RunnableProxy implements InvocationHandler {
+
+    private final String globalTxId;
+    private final String localTxId;
+    private final String parentTxId;
+    private final Object runnable;
```
WillemJiang: If it is Runnable, we can just treat it as Runnable.

seanyinx: yeah, we can. but why do the casting?

WillemJiang: OK,  so we use Object to support Runnable and Callable this two cases.

PR #105, opened 2018-01-06
SCB-174 add kryo serializer

eric-lee-ltk:
Signed-off-by: Eric Lee <dagang.li@huawei.com>

---


pom.xml
```diff
@@ -329,6 +329,11 @@
         <artifactId>grpc-stub</artifactId>
         <version>${grpc.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.esotericsoftware</groupId>
+        <artifactId>kryo</artifactId>
+        <version>4.0.1</version>
```
WillemJiang: How about using the ${kryo.version}?

omega/omega-format/src/test/java/org/apache/servicecomb/saga/omega/format/MessageFormatTestBase.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.startsWith;
+import static org.junit.Assert.assertThat;
+
+import org.apache.servicecomb.saga.omega.transaction.OmegaException;
+import org.apache.servicecomb.saga.omega.transaction.TxEvent;
+
+public class MessageFormatTestBase {
```
WillemJiang: You can use @Ignore annotation to avoid running the test in base class, and add  Test annotation  in base  class.

omega/omega-format/src/test/java/org/apache/servicecomb/saga/omega/format/KryoMessageFormatTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class KryoMessageFormatTest extends MessageFormatTestBase {
+
+  @BeforeClass
+  public static void setUp() {
+    format = new KryoMessageFormat();
+  }
+
+  @Test
+  @Override
```
WillemJiang: We can remove these lines when putting the Test annotation to the base class.

omega/omega-format/src/main/java/org/apache/servicecomb/saga/omega/format/NativeMessageFormat.java
```diff
@@ -23,21 +23,9 @@
...
-  @Override
-  public byte[] serialize(TxEvent event) {
-    try {
-      return serialize(event.payloads());
-    } catch (OmegaException e) {
-      throw new OmegaException("Unable to serialize event with global tx id " + event.globalTxId(), e);
-    }
-  }
-
+public class NativeMessageFormat implements MessageFormat {
```
seanyinx: `NativeMessageFormat` can be removed

omega/omega-format/src/main/java/org/apache/servicecomb/saga/omega/format/KryoMessageFormat.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final int DEFAULT_BUFFER_SIZE = 4096;
+
+  private static final KryoFactory factory = Kryo::new;
+
+  private static final KryoPool pool = new KryoPool.Builder(factory).softReferences().build();
+
+  @Override
+  public byte[] serialize(Object[] objects) {
+    Output output = new Output(DEFAULT_BUFFER_SIZE, -1);
```
seanyinx: what happens if serialized bytes are larger than default buffer size?

eric-lee-ltk: It will expand twice large of the current buffer size until it fits the actual size as we set the maximum to be -1 which means unlimited.

PR #107, opened 2018-01-08
[SCB-168] load balance

seanyinx:

---

### DIFF
--- a/omega/omega-connector/omega-connector-grpc/pom.xml
+++ b/omega/omega-connector/omega-connector-grpc/pom.xml
@@ -52,6 +52,18 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+    </dependency>
 
     <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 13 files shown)


omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
```diff
@@ -0,0 +1,140 @@
...
+  public void close() {
+    channels.forEach(ManagedChannel::shutdownNow);
+  }
+
+  @Override
+  public void send(TxEvent event) {
+    boolean success = false;
+    do {
+      try {
+        withFastestSender(messageSender -> {
```
WillemJiang: I think we should add sleep for it if the message is not sent out rightly.

seanyinx: doing so will slow down the entire transaction...

omega/omega-connector/omega-connector-grpc/pom.xml
```diff
@@ -53,6 +53,21 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
```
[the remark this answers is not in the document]
seanyinx: this is for test only

omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
WillemJiang: If the sender always failed to send  the message,  the CPU usage could be very high.

seanyinx: that's what https://issues.apache.org/jira/browse/SCB-211 is for due to entire cluster unreachable

PR #111, opened 2018-01-10
SCB-213 proper timing to handle SagaEndedEvent in asynchronous situation

eric-lee-ltk:
Signed-off-by: Eric Lee <dagang.li@huawei.com>

---

### DIFF
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
@@ -24,12 +24,12 @@ import static org.apache.servicecomb.saga.alpha.core.EventType.TxStartedEvent;
 
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.function.Consumer;
 
 public class TxConsistentService {
@@ -61,8 +61,8 @@ public class TxConsistentService {
     List<TxEvent> events = eventRepository.findStartedTransactions(event.globalTxId(), TxStartedEvent.name());
     events.forEach(omegaCallback::compensate);
     eventsToCompensate.computeIfAbsent(event.globalTxId(), (v) -> {
-      CopyOnWriteArraySet<String> eventSet = new CopyOnWriteArraySet<>();
-      events.forEach(e -> eventSet.add(getUniqueEventId(e)));
+      Set<String> eventSet = new HashSet<>(events.size());
+      events.forEach(e -> eventSet.add(e.localTxId()));
       return eventSet;
     });
   }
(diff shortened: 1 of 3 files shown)


alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
```diff
@@ -44,7 +58,32 @@ public void handle(TxEvent event) {
   }
 
   private void compensate(TxEvent event) {
-    List<TxEvent> events = eventRepository.findStartedTransactions(event.globalTxId(), EventType.TxStartedEvent.name());
+    List<TxEvent> events = eventRepository.findStartedTransactions(event.globalTxId(), TxStartedEvent.name());
     events.forEach(omegaCallback::compensate);
+    eventsToCompensate.computeIfAbsent(event.globalTxId(), (v) -> {
+      CopyOnWriteArraySet<String> eventSet = new CopyOnWriteArraySet<>();
+      events.forEach(e -> eventSet.add(getUniqueEventId(e)));
```
seanyinx: what's the performance characteristics of CopyOnWrite?

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
```diff
@@ -44,7 +58,32 @@ public void handle(TxEvent event) {
...
+    if (events != null) {
+      events.remove(getUniqueEventId(event));
+      if (events.isEmpty()) {
+        markGlobalTxEnd(event);
+      }
+    }
+  }
+
+  private String getUniqueEventId(TxEvent event) {
+    return event.globalTxId() + "_" + event.localTxId();
```
seanyinx: the map is based on globalTxId. why use globalTxId again?

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
eric-lee-ltk: There is no need to use CopyOnWriteArraySet as the set is only write once. Already used HashSet instead.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
eric-lee-ltk: That's a good point. Use localTxId is enough.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java
```diff
@@ -44,7 +58,28 @@ public void handle(TxEvent event) {
...
+      return eventSet;
+    });
+  }
+
+  private void updateCompensateStatus(TxEvent event) {
+    Set<String> events = eventsToCompensate.get(event.globalTxId());
+    if (events != null) {
+      events.remove(event.localTxId());
+      if (events.isEmpty()) {
+        markGlobalTxEnd(event);
```
seanyinx: how about remove the set from map, or it will keep growing?

PR #112, opened 2018-01-11
[SCB-212] tx timeout

seanyinx:
Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/SCB) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[SCB-XXX] Fixes bug in ApproximateQuantiles`, where you replace `SCB-XXX` with the appropriate JIRA issue.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean install` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

---

PR #113, opened 2018-01-11
SCB-211 wait until connection recover instead of try sending

eric-lee-ltk:

---

### DIFF
--- a/omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
+++ b/omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
@@ -24,15 +24,13 @@ import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.locks.Condition;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 
 import org.apache.servicecomb.saga.omega.context.ServiceConfig;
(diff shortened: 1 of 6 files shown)


eric-lee-ltk:
This PR is based on #112

omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
```diff
@@ -139,10 +145,27 @@ public void send(TxEvent event) {
...
     } while (!success && !Thread.currentThread().isInterrupted());
   }
 
+  private void waitTillAnyConnectionAvailable() {
+    MessageSender sender = fastestSender();
+    // no connection available
+    if (senders.get(sender) == Long.MAX_VALUE) {
+      lock.lock();
+      try {
+        condition.await();
```
WillemJiang: Do we have timeout off it?

omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/PushBackReconnectRunnable.java
```diff
@@ -31,26 +33,39 @@
...
     this.messageSender = messageSender;
     this.senders = senders;
     this.pendingTasks = pendingTasks;
+    this.lock = lock;
+    this.condition = condition;
   }
 
   @Override
   public void run() {
+    lock.lock();
```
WillemJiang: why do we need the lock here?

omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/LoadBalancedClusterMessageSender.java
```diff
@@ -50,6 +52,10 @@
   private final Map<MessageSender, Long> senders = new HashMap<>();
   private final Collection<ManagedChannel> channels;
 
+  private final ReentrantLock lock = new ReentrantLock();
+
+  private final Condition condition = lock.newCondition();
```
seanyinx: locks are hard to reason and easy to problems. not recommended to use them.

zhfeng:
As I understand that the MessageSender.send( ) will be blocked until the backend alpha servers become available. So in the situation of the all of the alpha servers are down for the outage , the SagaStart will be hang. I just wonder if there is a way for throwing an exception to say "Can not start Saga" ?

zhfeng:
The orgin [SCB-211](https://issues.apache.org/jira/browse/SCB-211) suppose to do the exponential backoff re-connecting.  This article [Retrying-transactions-with-exponentia](http://blog.astithas.com/2009/06/retrying-transactions-with-exponential.html) could be helpful.

seanyinx:
@zhfeng yes, when none of the alphas are reachable at @SagaStart, we shall immediately fail the transaction. we can raise a PR for that, when apache jira is online.

PR #116, opened 2018-01-13
SCB-220 avoid redundant compensation

seanyinx:

---

### DIFF
--- a/alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/TxEventEnvelopeRepository.java
+++ b/alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/TxEventEnvelopeRepository.java
@@ -19,15 +19,33 @@ package org.apache.servicecomb.saga.alpha.server;
 
 import java.util.List;
 
+import org.apache.servicecomb.saga.alpha.core.TxEvent;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
 interface TxEventEnvelopeRepository extends CrudRepository<TxEventEnvelope, Long> {
-  TxEventEnvelope findByEventGlobalTxId(String globalTxId);
+  List<TxEventEnvelope> findByEventGlobalTxId(String globalTxId);
 
-  @Query("SELECT DISTINCT new org.apache.servicecomb.saga.alpha.server.TxEventEnvelope("
+  @Query("SELECT DISTINCT new org.apache.servicecomb.saga.alpha.core.TxEvent("
       + "t.event.serviceName, t.event.instanceId, t.event.globalTxId, t.event.localTxId, t.event.parentTxId, t.event.type, t.event.compensationMethod, t.event.payloads"
       + ") FROM TxEventEnvelope t "
       + "WHERE t.event.globalTxId = ?1 AND t.event.type = ?2")
-  List<TxEventEnvelope> findByEventGlobalTxIdAndEventType(String globalTxId, String type);
+  List<TxEvent> findByEventGlobalTxIdAndEventType(String globalTxId, String type);
+
+  TxEventEnvelope findFirstByEventGlobalTxIdAndEventLocalTxIdAndEventType(String globalTxId, String localTxId, String type);
+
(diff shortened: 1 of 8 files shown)


seanyinx:
on top of https://github.com/apache/incubator-servicecomb-saga/pull/112

alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/TxEventEnvelopeRepository.java
```diff
@@ -19,15 +19,33 @@
...
+      + "  FROM TxEventEnvelope t1 "
+      + "  WHERE t1.event.globalTxId = ?1 "
+      + "  AND t1.event.localTxId = t.event.localTxId "
+      + "  AND t1.event.type = 'TxEndedEvent'"
+      + ") AND NOT EXISTS ( "
+      + "  FROM TxEventEnvelope t2 "
+      + "  WHERE t2.event.globalTxId = ?1 "
+      + "  AND t2.event.localTxId = t.event.localTxId "
+      + "  AND t2.event.type = 'TxCompensatedEvent')")
+  List<TxEvent> findStartedEventsWithMatchingEndedButNotCompensatedEvents(String globalTxId);
```
eric-lee-ltk: seems the started events are not necessary since you just need to find out all ended events without their corresponding compensated events?

seanyinx: only started event contains payloads, that's why we need to search for them

eric-lee-ltk: How about using the group by and having similar to the following statement?
```sql
SELECT xxx FROM TxEventEnvelope t LEFT JOIN (SELECT globalTxId FROM TxEventEnvelope WHERE type = 'TxStartedEvent' OR type = 'TxEndedEvent' OR type = 'TxCompensatedEvent' GROUP BY globalTxId HAVING COUNT(*) = 2) AS t1 ON t.globalTxId = t1.globalTxId;
```

seanyinx: did you try this query? left join returns all records from left table...

eric-lee-ltk: No. I just suggest there maybe another way to do this job using only two queries instead of three. As you see, the second sub query is quite complicate and I have no idea whether the performance of this statement is optimal. Besides, I missed the started event filter of  the first query part, sorry.

seanyinx: one of the sub queries in exists and the other is not exists. i am not sure how that can be done in a single join...

eric-lee-ltk: Suppose there are two queries:
1. query that retrieves all the StartedEvent with a given globalTxId
2. query that retrieves events of either Started, Ended or Compensated of a given globalTxId, events are group by the localTxId and we can filter compensated events based on the count(2, either Started or Ended) of each group using the `having`. 

Finally, we join the above queries to get the events that need to compensate.

seanyinx: i don't quite understand. there may be duplicate started event, and how to make sure count(2) to exclude compensated event?

eric-lee-ltk: duplicate started event with the same localTxId? Does this happen when the omega recovers?

seanyinx: maybe. and also possible when network unstable

PR #118, opened 2018-01-14
SCB-232 EventType should be same between omega and alpha

zhfeng:

---

https://issues.apache.org/jira/browse/SCB-232

### DIFF
--- a/omega/omega-transaction/pom.xml
+++ b/omega/omega-transaction/pom.xml
@@ -32,6 +32,10 @@
   <dependencies>
     <dependency>
       <groupId>org.apache.servicecomb.saga</groupId>
+      <artifactId>pack-common</artifactId>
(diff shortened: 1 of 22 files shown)


omega/omega-transaction/pom.xml
```diff
@@ -30,6 +30,10 @@
   <name>Saga::Omega::Transaction</name>
 
   <dependencies>
+    <dependency>
```
seanyinx: if we really want to share code between omega and alpha, is it better to do it in a separate module?

zhfeng: yeah, I agree. How about the "common" module ?

seanyinx: ok, pack-common sounds good

PR #122, opened 2018-01-19
SCB-239 omega stateless

eric-lee-ltk:

---


alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/GrpcTxEventEndpointImpl.java
```diff
@@ -54,14 +58,22 @@
...
+        .put(request.getInstanceId(), new GrpcOmegaCallback(responseObserver));
+
+    txConsistentService.recompensateByInstanceId(request.getInstanceId());
   }
 
   // TODO: 2018/1/5 connect is async and disconnect is sync, meaning callback may not be registered on disconnected
   @Override
   public void onDisconnected(GrpcServiceConfig request, StreamObserver<GrpcAck> responseObserver) {
+    LOG.warn("Abort all started event of instance {} because connection is closing.", request.getInstanceId());
+    txConsistentService.abortAllStartedEventsByInstanceId(request.getServiceName(), request.getInstanceId());
```
seanyinx: why abort all transactions when an instance is down?

eric-lee-ltk: When the SagaStart node fails, the transaction is no longer valid because user will get an error immediately instead of the subsequent result. We should stop it from executing.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -64,10 +66,12 @@ public void run() {
   private void pollEvents() {
     scheduler.scheduleWithFixedDelay(
         () -> {
+          abortTimeoutEvent();
           saveUncompensatedEventsToCommands();
           compensate();
           updateCompensatedCommands();
-          deleteDuplicateSagaEndedEvents();
+          deleteDuplicateEvents();
```
WillemJiang: Why do you have the duplicated events ?  The scheduler work should be done without any interruption, what if the delete work is finished  but update transaction status work is failed.

eric-lee-ltk: Multiple alphas handling the same event will create the duplicates. The situation you mention will be  handled at the next execution of scheduler or another alpha since all status are stored in db.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -92,11 +96,11 @@ private void updateCompensatedCommands() {
         });
   }
 
-  private void deleteDuplicateSagaEndedEvents() {
+  private void deleteDuplicateEvents() {
     try {
-      eventRepository.deleteDuplicateEvents(SagaEndedEvent.name());
+      eventRepository.deleteDuplicateEvents(Arrays.asList(TxAbortedEvent.name(), SagaEndedEvent.name()));
```
seanyinx: we don't delete duplicate events other than SagaEndedEvent. we need to keep track of all events for troubleshooting and auditing purpose.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -120,16 +137,29 @@ private void markGlobalTxEnd(TxEvent event) {
...
+    return new TxEvent(
+        event.serviceName(),
+        event.instanceId(),
+        event.globalTxId(),
+        event.localTxId(),
+        event.parentTxId(),
+        TxAbortedEvent.name(),
+        "",
+        null,
+        EMPTY_PAYLOAD);
```
seanyinx: please specify the reason for abort

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -109,6 +113,19 @@ private void updateCompensationStatus(TxEvent event) {
...
+    eventRepository.findFirstTimeoutEventByIdGreaterThan(nextTimeoutEventId)
+    .ifPresent((TxEvent event) -> {
+      log.info("Found timeout event {}", event);
+      nextTimeoutEventId = event.id();
+      eventRepository.save(toTxAbortedEvent(event));
+    });
+  }
+
+  private void updateTransactionStatus() {
+    eventRepository.findFirstAbortedGlobalTransaction().ifPresent(this::markGlobalTxEnd);
```
seanyinx: why mark end without any compensation?

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxEvent.java
```diff
@@ -64,21 +66,25 @@ public TxEvent(
...
       String globalTxId,
       String localTxId,
       String parentTxId,
       String type,
       String compensationMethod,
+      int timeout,
       byte[] payloads) {
-    this(id, serviceName, instanceId, new Date(), globalTxId, localTxId, parentTxId, type, compensationMethod, payloads);
+    this(surrogateId, serviceName, instanceId, creationTime, globalTxId, localTxId, parentTxId, type,
+        compensationMethod, timeout == 0 ? null : new Date(creationTime.getTime() + timeout*1000), payloads);
```
seanyinx: why multiply by 1000 here, but not below? prefer not to deal with null, give it a big timeout instead, e.g. 1000 years.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
seanyinx: why not lock event row when scanning for timeout event to avoid duplicate aborted event?

omega/omega-transaction/src/main/java/org/apache/servicecomb/saga/omega/transaction/SagaStartAspect.java
```diff
@@ -51,11 +47,10 @@ Object advise(ProceedingJoinPoint joinPoint, SagaStart sagaStart) throws Throwab
     initializeOmegaContext();
     Method method = ((MethodSignature) joinPoint.getSignature()).getMethod();
 
-    TimeAwareInterceptor interceptor = new TimeAwareInterceptor(sagaStartAnnotationProcessor);
-    interceptor.preIntercept(context.globalTxId(), method.toString());
+    OnceAwareInterceptor interceptor = new OnceAwareInterceptor(sagaStartAnnotationProcessor);
```
seanyinx: we don't need to wrap sagaStartAnnotationProcessor if no scheduled timeout task

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -109,6 +113,19 @@ private void updateCompensationStatus(TxEvent event) {
     markSagaEnded(event);
   }
 
+  private void abortTimeoutEvent() {
+    eventRepository.findFirstTimeoutEventByIdGreaterThan(nextTimeoutEventId)
+    .ifPresent((TxEvent event) -> {
+      log.info("Found timeout event {}", event);
+      nextTimeoutEventId = event.id();
+      eventRepository.save(toTxAbortedEvent(event));
```
seanyinx: we need to compensate timed out event too, since there's no way to know if its sub tx was done

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
eric-lee-ltk: timeout event with only SagaStartedEvent or only SagaStartedEvent and TxStartedEvent doesn't have anything to compensate, but currently there's no way to mark the end of such tx

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxEvent.java
eric-lee-ltk: the timeout is in second while the getTime is in milliseconds, that's why we multiply by 1000. You mean we should prefer to set the expireTime to be 9999-12-31 11:59:59 as the default value in db instead of null？

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
eric-lee-ltk: the lock will be invalid during the query of one alpha, what if another alpha visit the same row after the lock released?

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxEvent.java
seanyinx: yes, doing so can keep our logic simpler

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
seanyinx: we can check if such a event already exists. as long as there's no concurrent processing, it has much less chance running into duplication

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -109,6 +113,19 @@ private void updateCompensationStatus(TxEvent event) {
     markSagaEnded(event);
   }
 
+  private void abortTimeoutEvent() {
+    eventRepository.findFirstTimeoutEventByIdGreaterThan(nextTimeoutEventId)
```
seanyinx: checking timeout cannot be done by scanning through event ids, because some to-be-expired event will be skipped.

WillemJiang:
@eric-lee-ltk  Could you the rebase work to resolve the conflicts?

alpha/alpha-server/src/main/resources/schema-postgresql.sql
```diff
@@ -30,3 +31,15 @@ CREATE TABLE IF NOT EXISTS Command (
 );
 
 CREATE INDEX IF NOT EXISTS saga_commands_index ON Command (surrogateId, eventId, globalTxId, localTxId, status);
+
+
+CREATE TABLE IF NOT EXISTS TxTimeout (
+  surrogateId BIGSERIAL PRIMARY KEY,
+  globalTxId varchar(36) NOT NULL,
+  localTxId varchar(36) NOT NULL,
+  expireTime TIMESTAMP NOT NULL,
```
seanyinx: expiryTime

alpha/alpha-core/src/test/java/org/apache/servicecomb/saga/alpha/core/TxConsistentServiceTest.java
```diff
@@ -70,6 +78,29 @@ public void deleteDuplicateEvents(String type) {
...
+        if (timeout.globalTxId().equals(globalTxId) && timeout.localTxId().equals(localTxId)) {
+          timeout.setStatus(DONE.name());
+          break;
+        }
+      }
+    }
+
+    @Override
+    public List<TxEvent> findFirstTimeoutTxToAbort() {
+      return null;
```
seanyinx: empty list instead of null

alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/TxTimeoutEntityRepository.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Lock(LockModeType.OPTIMISTIC)
+  @Query("SELECT te FROM TxEvent AS te "
+      + "INNER JOIN TxTimeout AS tt "
+      + "ON te.globalTxId = tt.globalTxId "
+      + "  AND te.localTxId = tt.localTxId "
+      + "  AND tt.status = 'NEW' "
+      + "  AND tt.expireTime < CURRENT_TIMESTAMP "
+      + "ORDER BY tt.expireTime ASC")
+  List<TxEvent> findFirstTimeoutTxOrderByExpireTimeAsc(Pageable pageable);
```
seanyinx: try to copy everything you need to timeout table and avoid locking event table

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -64,16 +67,32 @@ public void run() {
...
         0,
         eventPollingInterval,
         MILLISECONDS);
   }
 
+  private void findTimeoutEvents() {
+    eventRepository.findTimeoutEvents()
+        .forEach(event -> {
+          log.info("Found timeout event {}", event);
+          timeoutRepository.save(txTimeoutOf(event));
```
seanyinx: will this create duplicate timeout entries when multiple alphas work on the same event??

eric-lee-ltk: Yes, it will throw the integrity constraint violation exception when other alphas save the same record in db just like the command behaves.

PR #154, opened 2018-03-14
[SCB-260] Add third party notices

yangbor:

---

### DIFF
--- a/saga-distribution/src/release/NOTICE
+++ b/saga-distribution/src/release/NOTICE
@@ -3,3 +3,25 @@ Copyright 2017-2018 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
+
+
+================================================================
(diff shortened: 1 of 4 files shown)


saga-distribution/src/release/NOTICE
```diff
@@ -3,3 +3,19 @@ Copyright 2017-2018 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
+
+
+================================================================
+Notice for gRPC
+
+See licenses/NOTICE-grpc
```
WillemJiang: It's better to use the notices director instead of licenses.

yangbor: From my POV it's irrelevent where we put the NOTICE-xxx files so long as the end user can get those files. And put them under the licenses folder have the advantage that the end user can get all the legal information from one place.

PR #157, opened 2018-03-22
[SCB-426] change type of CompensationContext.contexts to ConcurrentHashMap

chanjarster:
change type of CompensationContext.contexts to ConcurrentHashMap


seanyinx:
@chanjarster why this hash map needs to be concurrent?

chanjarster:
@seanyinx ``CompensableAnnotationProcessor``, which is a ``BeanPostProcessor``, uses ``CompensableMethodCheckingCallback`` in method `checkMethod`:

```java
  private void checkMethod(Object bean) {
    ReflectionUtils.doWithMethods(
        bean.getClass(),
        new CompensableMethodCheckingCallback(bean, compensationContext));
  }
``` 

Which in turn registers compensation method to `CompensationContext`:

```java
```

PR #173, opened 2018-04-16
[SCB-300] acceptance test for @Compensable event timeout

lijasonvip:

---
Now we have acceptance test for `@SagaStart` event timeout in #171 , we need to test `@Compensable` event timeout too.

### DIFF
--- a/acceptance-tests/acceptance-pack/src/test/resources/hotel_timeout.btm
+++ b/acceptance-tests/acceptance-pack/src/test/resources/hotel_timeout.btm
@@ -1,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+##############################################################
+# rules to timeout the car after invoking the services
+#
+###############################################################
+RULE set the saga compensable timeout to 2s
+INTERFACE org.apache.servicecomb.saga.omega.transaction.annotations.Compensable
+METHOD timeout
+AT EXIT
+IF TRUE
+DO RETURN 2
+ENDRULE
+
+RULE sleep when postHotelBooking until timeout happens
+CLASS org.apache.servicecomb.saga.demo.pack.hotel.HotelBookingService
+METHOD postHotelBooking
+AT ENTRY
+IF TRUE
+DO debug("delay 10s until the car booking timeout"),
+   Thread.sleep(4000)
+ENDRULE
--- a/acceptance-tests/acceptance-pack/src/test/resources/pack_car_timeout_scenario.feature
+++ b/acceptance-tests/acceptance-pack/src/test/resources/pack_car_timeout_scenario.feature
@@ -1,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


acceptance-tests/acceptance-pack/src/test/resources/hotel_timeout.btm
```diff
@@ -0,0 +1,35 @@
...
+IF TRUE
+DO RETURN 2
+ENDRULE
+
+RULE sleep when postHotelBooking until timeout happens
+CLASS org.apache.servicecomb.saga.demo.pack.hotel.HotelBookingService
+METHOD postHotelBooking
+AT ENTRY
+IF TRUE
+DO debug("delay 10s until the car booking timeout"),
```
WillemJiang: Debug log is not right

acceptance-tests/acceptance-pack/src/test/resources/pack_hotel_timeout_scenario.feature
```diff
@@ -0,0 +1,47 @@
...
+    Then Alpha records the following events
+      | serviceName  | type               |
+      | pack-booking | SagaStartedEvent   |
+      | pack-car     | TxStartedEvent     |
+      | pack-car     | TxEndedEvent       |
+      | pack-hotel   | TxStartedEvent     |
+      | pack-hotel   | TxAbortedEvent     |
+      | pack-car     | TxCompensatedEvent |
+      | pack-hotel   | TxCompensatedEvent |
+      | pack-hotel   | SagaEndedEvent     |
```
WillemJiang: I don't think pack-hotel should send the SagaEndedEvent here.

acceptance-tests/acceptance-pack/src/test/resources/pack_hotel_timeout_scenario.feature
```diff
@@ -0,0 +1,47 @@
...
+      | pack-hotel   | TxEndedEvent       |
+      | pack-booking | SagaEndedEvent     |
+
+    Then Car Service contains the following booking orders
+      | name  | amount | confirmed | cancelled |
+      | Jason | 1      | false     | true      |
+
+    And Hotel Service contains the following booking orders
+      | name  | amount | confirmed | cancelled |
+      | Jason | 1      | true      | false     |
```
WillemJiang: The Hotel Service should be cancelled.  There must be some errors in the Alpha Service.

acceptance-tests/acceptance-pack/src/test/resources/pack_car_timeout_scenario.feature
```diff
@@ -0,0 +1,39 @@
...
+      | pack-booking | SagaStartedEvent   |
+      | pack-car     | TxStartedEvent     |
+      | pack-car     | TxAbortedEvent     |
+      | pack-car     | TxCompensatedEvent |
+      | pack-car     | SagaEndedEvent     |
+      | pack-car     | TxEndedEvent       |
+
+    Then Car Service contains the following booking orders
+      | name  | amount | confirmed | cancelled |
+      | Jason | 1      | true      | false     |
```
WillemJiang: The Car Service should be cancelled.

WillemJiang:
It looks like the Accept test result is not good. We need to fix the issue here.

zhfeng:
what is the purpose of this test ? I think it would like to show the scenario that when the saga transaction is timeout, the alpha server call the compensate method is failed and will retry ?

The booking service looks like the following steps:
1) saga start
2) call the car booking
3) call the hotel booking
4) postBooking wait 10s until the transaction timeout (injected by the byteman rules)
5) return failure message with the information of timeout ( I think it is still an issue when we are sending the SAGA_ENDED in the https://github.com/apache/incubator-servicecomb-saga/blob/master/omega/omega-transaction/src/main/java/org/apache/servicecomb/saga/omega/transaction/SagaStartAnnotationProcessor.java#L45). I think we should check the return value and throws the Exception if the transaction is timeout. Also the alpha server handles the message in (https://github.com/apache/incubator-servicecomb-saga/blob/master/alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxConsistentService.java#L38), currently it only check if the transaction is aborted and I think it should also check if the transaction is timeout.

For this test, if we want to simulate the alpha server retries the compensate method when the transaction is timeout, I think it would like to use the byteman rule in the cancel method of HotelBooking (https://github.com/apache/incubator-servicecomb-saga/blob/master/saga-demo/booking/hotel/src/main/java/org/apache/servicecomb/saga/demo/pack/hotel/HotelBookingService.java#L40)
```
RULE create counter
CLASS org.apache.servicecomb.saga.demo.pack.hotel.HotelBooking
METHOD <init>
IF TRUE
DO createCountDown($0, 1)
ENDRULE

RULE throws the exception when calling the cancel method at the first time
CLASS org.apache.servicecomb.saga.demo.pack.hotel.HotelBookingService
MEHTOD cancel(HotelBooking)
AT ENTRY
IF !countDown($0)
DO throw exception
ENDRULE
```

acceptance-tests/acceptance-pack/src/test/resources/hotel_timeout.btm
lijasonvip: My bad.

acceptance-tests/acceptance-pack/src/test/resources/pack_hotel_timeout_scenario.feature
lijasonvip: I'll fix this.

PR #215, opened 2018-07-11
SCB-732 Add the omega-transport-feign into distribution kit and updated LICENSE 

Lionelpang:
1. Add the omega-transport-feign into distribution kit
2. Add the Openfeign LICENSE to the release/licenses fold.


WillemJiang:
Thx for the contribution.

PR #255, opened 2018-08-20
SCB-852 Create saga-core-akka module

WillemJiang:
Split akka related code to saga-core-akka


WillemJiang:
After Running the build on my Mac, all test passed.

PR #258, opened 2018-08-21
[SCB-855]Support JDK7 in saga-format

KomachiSion:
Rewrite the lamda function in saga-format to support JDK 7


saga-format/src/test/java/org/apache/servicecomb/saga/format/JacksonFromJsonFormatTest.java
```diff
@@ -166,36 +195,37 @@ public void setUp() throws Exception {
         .thenReturn(response11);
     when(restTransport.with("aaa", "/rest/as", "delete", singletonMap("query", singletonMap("bar", "as"))))
         .thenReturn(response12);
-    when(restTransport.with("aaa", "/rest/as", "put", emptyMap()))
+    when(restTransport.with("aaa", "/rest/as", "put", Collections.<String, Map<String, String>>emptyMap()))
```
WillemJiang: We could extract an empty map constant here.

WillemJiang:
Please add JIRA number on your commit log, otherwise I had to use the squash to merge your change.

PR #262, opened 2018-08-21
SCB-865 Implement reaction of the event in Alpha Server

cherrylzhao:
Implement reaction of TCC event in alpha server, which involves
-  rpc OnConnected (GrpcServiceConfig) returns (stream GrpcTccCordinateCommand)
- rpc participate(GrpcTccParticipateEvent) returns (GrpcAck)
- rpc OnTccTransactionEnded (GrpcTccTransactionEndedEvent) returns (GrpcAck)
- rpc OnDisconnected (GrpcServiceConfig) returns (GrpcAck)


alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/tcc/server/event/ParticipateEvent.java
```diff
@@ -0,0 +1,113 @@
...
+ * Participate event.
+ *
+ * @author zhaojun
+ */
+public class ParticipateEvent {
+
+  private String globalTxId;
+  private String localTxId;
+  private String parentTxId;
+  private String serviceName;
```
WillemJiang: we don't need specify the serviceName and instanceId information here.

PR #270, opened 2018-08-24
[SCB-868] Add Kamon metrics to Alpha Server

maheshrajus:

---

### DIFF
--- a/alpha/alpha-server/src/main/resources/application.conf
+++ b/alpha/alpha-server/src/main/resources/application.conf
@@ -1,0 +1,88 @@
+## ---------------------------------------------------------------------------
+## Licensed to the Apache Software Foundation (ASF) under one or more
+## contributor license agreements.  See the NOTICE file distributed with
+## this work for additional information regarding copyright ownership.
+## The ASF licenses this file to You under the Apache License, Version 2.0
+## (the "License"); you may not use this file except in compliance with
+## the License.  You may obtain a copy of the License at
+##
+##      http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+## ---------------------------------------------------------------------------
+
+###############################
+# Kamon related configuration #
+###############################
+
+kamon {
+
+  metric {
+    filters {
+      trace.includes = ["**"]
+      trace-segment.includes = ["**"]
+      trace-segment.excludes = []
+      akka-actor.includes = []
+      akka-actor.excludes = ["**"]
+      akka-dispatcher.includes = []
+      akka-dispatcher.excludes = ["**"]
+    }
+  }
+
+  show-aspectj-missing-warning = no
+
+  modules {
+    kamon-annotation {
+      auto-start = no
+      requires-aspectj = yes
+    }
+
+    kamon-log-reporter {
+      auto-start = no
+      requires-aspectj = no
+    }
+
+    kamon-statsd {
+      auto-start = no
+      requires-aspectj = no
+    }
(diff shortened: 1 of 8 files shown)


alpha/pom.xml
```diff
@@ -35,6 +35,27 @@
...
+    <dependency>
+      <groupId>io.kamon</groupId>
+      <artifactId>kamon-core_2.12</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.kamon</groupId>
+      <artifactId>kamon-annotation_2.12</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.servicecomb.saga</groupId>
```
WillemJiang: alpha doesn't use saga-core , we should not add this dependency.

maheshrajus: ok.done

alpha/pom.xml
```diff
@@ -35,24 +35,21 @@
     <module>alpha-server</module>
   </modules>
 
+  <properties>
+    <kamon.version>0.6.0</kamon.version>
```
WillemJiang: We normally put the dependency properties into parent pom, so the other module could share the same properties.

alpha/pom.xml
```diff
@@ -35,24 +35,21 @@
...
+    <akka.version>2.3.12</akka.version>
+  </properties>
+
   <dependencies>
-    <dependency>
+   <dependency>
       <groupId>io.kamon</groupId>
-      <artifactId>kamon-core_2.12</artifactId>
+      <artifactId>kamon-core_2.11</artifactId>
```
WillemJiang: Why do you change the scala version from 2.12 to 2.11?

alpha/alpha-server/src/main/resources/application.conf
```diff
@@ -24,31 +24,37 @@ kamon {
...
+      akka-dispatcher.excludes = []
     }
   }
 
   show-aspectj-missing-warning = no
 
   modules {
     kamon-annotation {
-      auto-start = no
+      auto-start = yes
```
WillemJiang: Why did you change these settings?

alpha/pom.xml
```diff
@@ -75,20 +72,42 @@
...
-          <artifactId>kamon-autoweave_2.12</artifactId>
-          <version>0.6.5</version>
+          <artifactId>kamon-statsd_2.11</artifactId>
+          <version>${kamon.version}</version>
         </dependency>
         <dependency>
-          <groupId>org.aspectj</groupId>
-          <artifactId>aspectjweaver</artifactId>
+          <groupId>com.typesafe.akka</groupId>
+          <artifactId>akka-actor_2.11</artifactId>
```
WillemJiang: Current Alpha Server doesn't use Akka, why did you put these dependencies.

alpha/alpha-server/src/main/resources/application.conf
maheshrajus: while running the alpha server I have to provide auto-start yes for kamon.  I put here yes and forgot to revert. i will make by default "no" and when user required then they will pass argument "yes".

alpha/pom.xml
maheshrajus: saga-core-akka support is there right so i thought akka support is there. i will remove unwated dependency.

alpha/pom.xml
maheshrajus: kamon-autoweave  dependency 0.6.0 not available for 2.12 version. So i made all kamon dependency to 2.11 version so that all kamon classes should be in single version.

maheshrajus: I will modify to latest version which is 2.12 only

alpha/pom.xml
```diff
@@ -35,6 +35,21 @@
     <module>alpha-server</module>
   </modules>
 
+  <dependencies>
```
WillemJiang: We should put these dependencies into alpha-server.

WillemJiang:
Merged the patch by apply some changes on pom files with thanks to maheshrajus.

PR #271, opened 2018-08-25
[SCB-865] Refactoring the Omega Interceptors

imlijinting:
1.Remove the inheritance hierarchy between SagaStartAnnotationProcessor and EventAwareInterceptor.
2.Remove the NO_OP_INTERCEPTOR from the interface EventAwareInterceptor. Turning the NoOp Implementation into a class of static singleton pattern.


imlijinting:
Something wrong with the test cases. 
Close this.

PR #272, opened 2018-08-25
[SCB-865] Refactoring the Omega Interceptors

imlijinting:
1.Remove the inheritance hierarchy between SagaStartAnnotationProcessor and EventAwareInterceptor.
2.Remove the NO_OP_INTERCEPTOR from the interface EventAwareInterceptor. Turning the NoOp Implementation into a class of static singleton pattern.
3.Removing parentId when dealing with error.


omega/omega-transaction/src/main/java/org/apache/servicecomb/saga/omega/transaction/NoOpEventAwareInterceptor.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.servicecomb.saga.omega.transaction;
```
WillemJiang: We need to add the license header here.

WillemJiang:
@imlijinting  Don't worry, I just add the license header to the file.

PR #274, opened 2018-08-28
SCB-817, SCB-818  TCC Omega side implemenation

WillemJiang:
* Updated the TCC protocol for CoordinatedEvent
* Update the omega transactions code
* Implemented the TccEventServie of grpc

### DIFF
--- a/omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/GrpcTccEventService.java
+++ b/omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/GrpcTccEventService.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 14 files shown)


omega/omega-connector/omega-connector-grpc/src/main/java/org/apache/servicecomb/saga/omega/connector/grpc/GrpcTccEventService.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Override
+  public AlphaResponse tccTransactionStop(TccEndedEvent tccEndEvent) {
+    GrpcAck grpcAck = tccBlockingEventService.onTccTransactionEnded(convertTo(tccEndEvent));
+    return new AlphaResponse(grpcAck.getAborted());
+
+  }
+
+  @Override
+  public AlphaResponse coordinate(CoordinatedEvent coordinatedEvent) {
```
cherrylzhao: I can't understand coordinate method mean, when sending transactionEndEvent, alpha will send the cordinateCommand to omega instance asynchronous.
it is why onConnect method has a stream observer.

WillemJiang: It means alpha need to get the execution result of the coordinatedCommand.  This process can be asynchronized.

cherrylzhao: when received the coordinate feedback message, how alpha handle it，or just persistence the result.

WillemJiang: Alpha just need to persistent the event for management usage.

PR #277, opened 2018-08-29
Update the images' tag to 0.3.0-SNAPSHOT in dependency free trans demo

crystaldust:


WillemJiang:
We have the same issue in the dependency-free-transaction-demo.

PR #283, opened 2018-08-30
SCB-876 Implement of Omega TCC coordinate command callback and refactor

cherrylzhao:
Implement of Omega TCC coordinate command callback and refactor

### DIFF
--- a/omega/omega-context/src/main/java/org/apache/servicecomb/saga/omega/context/OmegaContext.java
+++ b/omega/omega-context/src/main/java/org/apache/servicecomb/saga/omega/context/OmegaContext.java
@@ -17,7 +17,8 @@
 
 package org.apache.servicecomb.saga.omega.context;
 
-import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * OmegaContext holds the globalTxId and localTxId which are used to build the invocation map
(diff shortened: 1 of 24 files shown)


omega/omega-spring-tx/src/test/java/org/apache/servicecomb/saga/omega/transaction/spring/TransactionInterceptionTest.java
```diff
@@ -362,6 +353,11 @@ public void passesOmegaContextAmongActors() throws Exception {
     actorSystem.terminate();
   }
 
+  @Test
+  public void tccWorkflowTest() {
```
WillemJiang: The test is empty.

omega/omega-context/src/main/java/org/apache/servicecomb/saga/omega/context/OmegaContext.java
```diff
@@ -27,6 +30,12 @@
   private final ThreadLocal<String> globalTxId = new InheritableThreadLocal<>();
   private final ThreadLocal<String> localTxId = new InheritableThreadLocal<>();
   private final IdGenerator<String> idGenerator;
+  private final ThreadLocal<Map<String, Object[]>> parameters = new InheritableThreadLocal<Map<String, Object[]>>() {
```
WillemJiang: The parameters could eat up the memory, and we need to remove it after the invocation is successed.
Don't worry, I will add a quick fix for it.

PR #286, opened 2018-09-03
[SCB-864] support multiple parameters for each sql

KomachiSion:

---

### DIFF
--- a/saga-core/src/main/java/org/apache/servicecomb/saga/core/SQLOperation.java
+++ b/saga-core/src/main/java/org/apache/servicecomb/saga/core/SQLOperation.java
@@ -23,18 +23,18 @@ import java.util.List;
 public class SQLOperation implements Operation {
 
   private final String sql;
-  private final List<String> params;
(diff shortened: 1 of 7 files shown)


saga-core/src/main/java/org/apache/servicecomb/saga/core/SQLOperation.java
```diff
@@ -23,18 +23,18 @@
 public class SQLOperation implements Operation {
 
   private final String sql;
-  private final List<String> params;
+  private final List<List<Object>> params;
 
-  public SQLOperation(String sql, List<String> params) {
+  public SQLOperation(String sql, List<List<Object>> params) {
```
WillemJiang: params are string type. Why do you change it to Object?

KomachiSion: params may be Integer, Date.  so I think Object may be more suitable than String?

PR #289, opened 2018-09-04
 [SCB-820] Demo of TCC model.

imlijinting:
[SCB-820] Demo of TCC model. 
---


WillemJiang:
FYI, the iCLA is received.

WillemJiang:
Merged the patch into master branch with thanks to Li jinting.

PR #291, opened 2018-09-08
SCB-856 Alpha TCC data persistence WIP

cherrylzhao:


alpha/alpha-server/src/main/resources/schema-postgresql.sql
```diff
@@ -68,3 +68,35 @@ CREATE TABLE IF NOT EXISTS TxTimeout (
 );
 
 CREATE INDEX IF NOT EXISTS saga_timeouts_index ON TxTimeout (surrogateId, expiryTime, globalTxId, localTxId, status);
+
+CREATE TABLE IF NOT EXISTS TccParticipateEvent (
```
WillemJiang: How about the TCC start event, we need to know it for timeout.

cherrylzhao: I haven't considered well about it.

WillemJiang:
Merged the code  into master branch by hand.

PR #299, opened 2018-09-17

oliugian:
https://issues.apache.org/jira/browse/SCB-915

do all compensation in one event scanner cycle,and if compensate failed,will try again.
fix out https://github.com/apache/incubator-servicecomb-saga/issues/253
optimize retry logic and judgement of tx abort event.


alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/Command.java
```diff
@@ -55,14 +55,14 @@
...
   private Command(long id,
-      String serviceName,
-      String instanceId,
-      String globalTxId,
-      String localTxId,
-      String parentTxId,
-      String compensationMethod,
-      byte[] payloads,
-      String status) {
+                  String serviceName,
```
WillemJiang: Please don't reformat the code that you don't change. 
You can setup your IDE to only reformat the code that VCS changed.
http://servicecomb.apache.org/developers/setup-develop-environment/

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/TxEventRepository.java
```diff
@@ -28,94 +28,69 @@
 
   /**
    * Save a {@link TxEvent}.
-   *
```
WillemJiang: These line should not be removed.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/CompositeOmegaCallbackRunner.java
```diff
@@ -20,30 +20,44 @@
...
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class CompositeOmegaCallback implements OmegaCallback {
+public class CompositeOmegaCallbackRunner implements OmegaCallback, Callable<List<TxEvent>> {
```
oliugian: Compensate Thread ， using by PushBackOmegaCallback . PushBackOmegaCallback  will filter all compensate event by there service name ,do all compensate event list with the same service name  in one  CompositeOmegaCallbackRunner  thread and only return those  compensate events which are successful.

oliugian: since omega call back is not thread safe. so compensate the same service use one thread.use other thread to compensate other service.

alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/EventScanner.java
```diff
@@ -79,58 +79,66 @@ private void pollEvents() {
     scheduler.scheduleWithFixedDelay(
         () -> {
           updateTimeoutStatus();
-          findTimeoutEvents();
+          findAllTimeoutEvents();
           abortTimeoutEvents();
           saveUncompensatedEventsToCommands();
           compensate();
           updateCompensatedCommands();
+          markSagaEndedForNoTxEnd();
```
oliugian: for these transaction: 
saga_start->tx_start->tx_abort->tx_abort 
we need to add a saga_end behind them ,so we can judge them finish.
for dumping finished transactions.

alpha/alpha-server/src/main/java/org/apache/servicecomb/saga/alpha/server/TxEventEnvelopeRepository.java
```diff
@@ -110,10 +111,65 @@
...
+      + "      AND t4.localTxId = t3.localTxId "
+      + "      AND t4.creationTime > t3.creationTime)) AND NOT EXISTS ( "
+      + "  SELECT t5.globalTxId FROM TxEvent t5 "
+      + "  WHERE t5.globalTxId = t.globalTxId "
+      + "    AND t5.localTxId = t.localTxId "
+      + "    AND t5.type = 'TxCompensatedEvent') AND NOT EXISTS ( "
+      + "   SELECT c FROM Command c "
+      + "   WHERE c.globalTxId = t.globalTxId "
+      + "    AND c.localTxId = t.localTxId ) ")
+  List<TxEvent> findNeedToCompensateTxs();
```
oliugian: for the try logic. if a transaction is nested,if this sub transaction is not done ,it must have a tx_start_event with no other event behind.
if a transaction is not nested.is's lasted tx event must be aborted.

PR #311, opened 2018-09-21

oliugian:
omega do not send every try event to alpha.omega do it's try logic by itself.


oliugian:
All tests have been successful
The functionality of ForwardRecovery has been incorporated into defaultRecovery, so delete it and its test code. There is a test code equivalent to ForwardRecover in TransactionAspectTest.
(-1) means infinite retry, but it can also be interrupted, throwing an InterruptedException, which is consistent with the logic in the previous ForwardRecovery

oliugian:
ignore it

PR #313, opened 2018-09-21

oliugian:
https://issues.apache.org/jira/browse/SCB-915


alpha/alpha-core/src/main/java/org/apache/servicecomb/saga/alpha/core/CompositeOmegaCallbackRunner.java
```diff
@@ -54,4 +68,37 @@ public void compensate(TxEvent event) {
...
+    for (TxEvent txEvent : txEvents) {
+      try {
+        LOG.info("compensating event with globalTxId: {} localTxId: {}", txEvent.globalTxId(),
+            txEvent.localTxId());
+        this.compensate(txEvent);
+        resultTxEvents.add(txEvent);
+      } catch (AlphaException ae) {
+        LOG.error("compensate event with globalTxId: {} localTxId: {} failed,error message is {}",
+            txEvent.globalTxId(), txEvent.localTxId(), ae);
+        break;
```
Chenhaoqing: I think it would be better to continue executing the looping.

PR #315, opened 2018-09-25
SCB-909 Add fail fast mechanism when all alpha is unreachable.

cherrylzhao:
Add fail fast mechanism when all alpha is unreachable, app can config it using omega.connection.sending.timeout


omega/omega-connector/omega-connector-grpc/src/test/java/org/apache/servicecomb/saga/omega/connector/grpc/saga/SagaLoadBalancedSenderTest.java
```diff
@@ -157,7 +157,11 @@ public void stopSendingOnInterruption() throws Exception {
     Thread thread = new Thread(new Runnable() {
       @Override
       public void run() {
-        messageSender.send(event);
+        try {
+          messageSender.send(event);
+        } catch (Exception ex) {
+          assertThat(ex.getMessage().endsWith("Failed to get reconnected sender"), is(true));
```
WillemJiang: The main thread cannot know if exception is thrown.

PR #322, opened 2018-10-22
[SCB-962] Support ForwardPolicy to configure the maximum of retries

KomachiSion:
Support ForwardPolicy to configure the maximum of retries

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -21,6 +21,7 @@ POST /requests
       "transaction": {
         "method": "",
         "path": "",
+        "retries": "",
         "params": {
 
         }
@@ -28,6 +29,7 @@ POST /requests
       "compensation": {
         "method": "",
         "path": "",
+        "retries": "",
         "params": {
 
         }
@@ -46,10 +48,12 @@ JSON parameters:
   - transaction - user-defined transaction that executed by the Saga.
     - method - user-defined, HTTP method.
     - path - user-defined, HTTP path.
+    - retries - int, optional, default -1. The max retry times for transaction. If this parameter is less than or equal to 0, transaction will retry infinitely.
     - params - support `form`,`json`,`body`,`query`.
   - compensation - user-defined compensation that executed by the Saga.
     - method - user-defined, HTTP method.
     - path - user-defined, HTTP path.
+    - retries - int, optional, default 3. The max retry times for compensation.
     - params - support `form`,`json`,`body`,`query`.
 
 2. Set content type to `text/plain`.
@@ -212,12 +216,14 @@ There are a `maven` example
       ],
       "transaction": {
         "sql": "",
+        "retries": "",
         "params": [
           []
         ]
       },
       "compensation": {
         "sql": "",
+        "retries": "",
         "params": [
           []
         ]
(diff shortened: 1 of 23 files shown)


docs/api/api.md
```diff
@@ -46,10 +48,12 @@ JSON parameters:
   - transaction - user-defined transaction that executed by the Saga.
     - method - user-defined, HTTP method.
     - path - user-defined, HTTP path.
+    - retries - int, optional, default -1. The max retry times for transaction. If this parameter is less than or equal to 0, transaction will retry infinitely.
     - params - support `form`,`json`,`body`,`query`.
   - compensation - user-defined compensation that executed by the Saga.
     - method - user-defined, HTTP method.
     - path - user-defined, HTTP path.
+    - retries - int, optional, default 3. The max retry times for compensation.
```
WillemJiang: Does it support -1 setting?

saga-core/src/main/java/org/apache/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -49,6 +49,15 @@ public SagaResponse apply(SagaTask task, SagaRequest request, SagaResponse paren
           ignored);
       throw new TransactionFailedException(ignored);
     }
+    throw new TransactionAbortedException(
+        String.format(
+            "Too many failures in transaction %s of service %s, stop transaction!",
```
WillemJiang: How about using "abort the transaction"?

docs/api/api.md
KomachiSion: When compensation set retries less than or equal to 0, it will be set default value.
This is the logic that was there, I have not changed it.

saga-core/src/main/java/org/apache/servicecomb/saga/core/Fallback.java
```diff
@@ -34,6 +34,11 @@ public SagaResponse send(String address) {
     public SagaResponse send(String address, SagaResponse response) {
       return send(address);
     }
+
+    @Override
+    public int retries() {
+      return 0;
```
zhfeng: why it return 0 here ? it means to retry infinite ?

saga-core/src/main/java/org/apache/servicecomb/saga/core/ForwardRecovery.java
```diff
@@ -49,6 +49,15 @@ public SagaResponse apply(SagaTask task, SagaRequest request, SagaResponse paren
...
     }
+    throw new TransactionAbortedException(
+        String.format(
+            "Too many failures in transaction %s of service %s, abort the transaction!",
+            request.transaction(),
+            request.serviceName()));
+  }
+
+  private boolean isRetryable(int i, Transaction transaction) {
+    return transaction.retries() == Transaction.INFINITE_RETRY || i <= transaction.retries();
```
zhfeng: I think it could be an issue here. if the transaction.retries() is 0, it just run sout of the loop and has no retry at all. But you mention in the document

> If this parameter is less than or equal to 0, transaction will retry infinitely.

zhfeng:
@KomachiSion Thanks for your contribution and can you check the travis-ci message at first. We have to make sure the all the tests pass.

saga-core/src/main/java/org/apache/servicecomb/saga/core/Fallback.java
KomachiSion: No, Fallback does not retry currently.
Because I abstracted the retires method from Compensation to Operation, Fallback needs to implement this method temporarily.
Considering the default number of retries if Fallback needs to retry in the future.

saga-core/src/main/java/org/apache/servicecomb/saga/core/ForwardRecovery.java
zhfeng: Thanks @KomachiSion , as i starts from 0
> for(int i = 0; isRetryable(i, request.transaction()); i++)
it looks like the condition should be i < transcaction.retries() ?

zhfeng:
@KomachiSion How to config the retries parameter globally as we discussed in the mailing list ? I think it could be possible to make the DEFAULT_RETRIES configurable ? Also can you squash these changes to one commit ?

KomachiSion:
I think it may not be reasonable to configure retries parameter globally. Because the retry features of transaction, compensation, and fallback may be inconsistent. If you need to configure global retries parameter, it may not be compatible with multiple cases.

About squashing commit, I execute as soon as possible.

zhfeng:
@KomachiSion Thanks for your contribution !

PR #325, opened 2018-10-30
SCB-995 Implement Saga Console API's

asifdxtreme:
This set of API's has been implemented to expose the saga events through API's which can be used by Saga Dashboard.
#### An Alpha-Server image with these changes are available [here](https://hub.docker.com/r/asifdiatm/alpha-server/) for test.

---


WillemJiang:
@asifdxtreme  Please add unit test for the code.

asifdxtreme:
> @asifdxtreme Please add unit test for the code. I also added some comments on the query API.

@WillemJiang thanks for the suggestion, I will add the duration parameter in the findTransaction API by which user can find the list of all transactions in happened in that duration.

One more thing, here in this complete set of API's I consider Global Transactions as 1 transaction and not the sub transactions, means if there is 5 Global transactions and each global transactions have 3 sub-transactions then the totalTransaction count as per my API will be 5 and not 15.

PR #332, opened 2018-11-09
SCB-973 update to remvoing the incubator

zhfeng:


zhfeng:
I just use the command to search in the plain text files.
```
find . -type f -exec grep -H 'incubator' '{}' \;
```

There are still two in the pom.xml and .travis.xml.setting which I'm not sure needing the changes. So maybe you can take a look.

PR #339, opened 2018-11-19
SCB-963 Add Test case for saga transactional compensation.

cherrylzhao:


omega/omega-spring-tx/src/test/java/org/apache/servicecomb/saga/omega/transaction/spring/TransactionalUserService.java
```diff
@@ -65,4 +65,17 @@ public void delete(User user, int count) {
...
+  public User addTransactional(User user) {
+    if (ILLEGAL_USER.equals(user.username())) {
+      throw new IllegalArgumentException("User is illegal");
+    }
+    return userRepository.save(user);
+  }
+
+  public void deleteTransactional(User user) {
+    userRepository.delete(user);
+    throw new RuntimeException("saga compensation rollback test");
```
WillemJiang: If the exception is thrown on purpose we need to change the name.

PR #375, opened 2019-01-11
change jpa.properties.eclipselink.ddl-generation in mysql profiles

cmonkey:
ddl-generation value none to create-or-extend-tables

---

### DIFF
--- a/alpha/alpha-server/src/main/resources/application.yaml
+++ b/alpha/alpha-server/src/main/resources/application.yaml
@@ -56,4 +56,4 @@ spring:
   jpa:
     properties:
       eclipselink:
-        ddl-generation: none
+        ddl-generation: create-or-extend-tables


alpha/alpha-server/src/main/resources/application.yaml
```diff
@@ -56,4 +56,4 @@ spring:
   jpa:
     properties:
       eclipselink:
-        ddl-generation: none
+        ddl-generation: create-or-extend-tables
```
WillemJiang: We could update the setting for postgresql and mysql just like this.
```
# EclipseLink will attempt to create tables. If the table exists, EclipseLink will add any missing columns.
spring:
  jpa:
    properties:
      eclipselink:
        ddl-generation: create-or-extend-tables

```

WillemJiang:
@cmonkey  Thanks for the PR, I just found this issue broke the demo when we upgrade the spring-boot version to 2.x.   My comments is just a change suggestion, as I need to fix the issue ASAP I just submit a PR #379 myself , you can take it as an example. Here is a [document](http://servicecomb.apache.org/developers/use-jira/) for how to use JIRA with github PR.  Looking forward your next  PR.

PR #413, opened 2019-02-21
SCB-1142 Clean up the alpha-spring-boot-starter

WillemJiang:
Just fix the version issue of alpha-spring-boot-starter.


coolbeevip:
> Just fix the version issue of alpha-spring-boot-starter.

Looks very clear

PR #415, opened 2019-02-25
[SCB-1171] Improve unit test coverage for ClusterLockServiceTest

coolbeevip:


alpha/alpha-server/src/test/java/org/apache/servicecomb/pack/alpha/server/cluster/master/ClusterLockServiceTest.java
```diff
@@ -60,33 +63,105 @@
...
+  @Test(timeout = 5000)
+  public void testMasterNode() throws InterruptedException {
+    MasterLock masterLock = clusterLockService.getMasterLock();
+    Assert.assertEquals(masterLock.getServiceName(),serviceName);
+    Assert.assertEquals(masterLock.getInstanceId(),instanceId);
+    Assert.assertEquals((masterLock.getExpireTime().getTime()-masterLock.getLockedTime().getTime()),expire);
+    when(masterLockEntityRepository.initLock(Mockito.anyString(),Mockito.any(Date.class),Mockito.any(Date.class),Mockito.anyString())).thenReturn(1);
+    when(masterLockEntityRepository.findMasterLockByServiceName(Mockito.anyString())).thenReturn(Optional.of(masterLock));
+    when(masterLockEntityRepository.updateLock(Mockito.anyString(),Mockito.any(Date.class),Mockito.any(Date.class),Mockito.anyString())).thenReturn(1);
+    Thread.sleep(2000);
```
WillemJiang: We can use waitUtil as we do in the accept test.

coolbeevip:
> @coolbeevip Can you update the test case with the
> await().atMost(2, SECONDS).until(() -> {
> xxx
> }
> to check the lock status?

OK

alpha/alpha-server/src/test/java/org/apache/servicecomb/pack/alpha/server/cluster/master/ClusterLockServiceTest.java
coolbeevip: OK, I will replace Mockito.any() with any()

alpha/alpha-server/src/test/java/org/apache/servicecomb/pack/alpha/server/cluster/master/ClusterLockServiceTest.java
```diff
@@ -60,37 +65,68 @@
...
+  private MasterLockEntityRepository masterLockEntityRepository;
 
-  @Test(timeout = 10000)
-  public void testClusterNodeType() throws InterruptedException {
-    //node type is master
-    clusterLockService.setMasterLock(null);
+  @Autowired
+  MasterLockRepository masterLockRepository;
+
+  @Before public void before(){
```
WillemJiang: It's strange for me to see you reset the test status.
If the masterLockEntityRepository is mocked, it's easy for us to reset it.
Is there anything that I'm missing?

coolbeevip: Use variable lockInitialization in the doLock method of AbstractLockProvider to avoid executing the initLock method again. InitLock method cannot be executed in unit test

WillemJiang: Thx for the explain, I will add some comment on the before method shortly.

PR #420, opened 2019-03-06
[SCB-1180] Register random gRPC port to eureka instance metadata

coolbeevip:


alpha/alpha-server/pom.xml
```diff
@@ -75,6 +75,10 @@
       <groupId>org.apache.servicecomb.pack</groupId>
       <artifactId>alpha-core</artifactId>
     </dependency>
+    <dependency>
```
WillemJiang: This change could introduce some trouble if we want to replace the service registry.

coolbeevip: If we support consul in the future, we can add the consul client dependency and then use which parameter control e.g

eureka.enabled=false
consul.enabled=true

alpha/alpha-spring-cloud-starter-eureka/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.servicecomb.pack</groupId>
+      <artifactId>alpha-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
```
WillemJiang: It should be in the test scope.

PR #424, opened 2019-03-09
[SCB-1190] Update the mysql manual for use the loader.path property to …

coolbeevip:
…define a lib folder location

Uses the `loader.path` system property and includes all JAR files in this directory and add them in the classpath

### DIFF
--- a/docs/faq/en/how_to_use_mysql_as_alpha_backend_database.md
+++ b/docs/faq/en/how_to_use_mysql_as_alpha_backend_database.md
@@ -1,46 +1,32 @@
 **Problem:** How to use MySQL as alpha's backend database?
 
 **Solution:** To replace the default database from postgreSQL to MySQL:
-1. add dependency of `mysql-connector-java` in `alpha/alpha-server/pom.xml`
-   ```xml
-       <dependency>
-         <groupId>mysql</groupId>
-         <artifactId>mysql-connector-java</artifactId>
-       </dependency>
-   ```
-
-2. install Saga
-   ```bash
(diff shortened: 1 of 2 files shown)


docs/faq/en/how_to_use_mysql_as_alpha_backend_database.md
```diff
@@ -1,46 +1,32 @@
...
+1. run MySQL
    ```bash
    docker run -d -e "MYSQL_ROOT_PASSWORD=password" -e "MYSQL_DATABASE=saga" -e "MYSQL_USER=saga" -e "MYSQL_PASSWORD=password" -p 3306:3306 mysql/mysql-server:5.7
    ```
 
-4. Run alpha. Please make sure MySQL is up before this step. You can run alpha through docker or executable file.
+2. Run alpha. Please make sure MySQL is up before this step. You can run alpha through docker or executable file.
    * via docker
+
+      Create `libs`  directory on the machine, for mount containers internal directory `/maven/saga/libs` , copy ` mysql-connector-java-8.0.15.jar` to the libs directory,  uses the `-Dloader.path=/maven/saga/libs` system property and includes all JAR files in this directory and add them in the classpath
```
WillemJiang: We need to tell user how to download the connector jar file.

coolbeevip: OK

coolbeevip: I added  a section about how to download mysql jdbc jar

PR #426, opened 2019-03-13
[SCB-1195] Modify eureka.enabled to eureka.client.enabled

coolbeevip:


coolbeevip:
> Why do we do this change? Can you explain it?

Wrong before,  eureka.client.enabled is the correct eureka parameter

PR #427, opened 2019-03-14
[SCB-1199] service discovery with consul

coolbeevip:
alpha register to consul

### DIFF
--- a/alpha/alpha-server/pom.xml
+++ b/alpha/alpha-server/pom.xml
@@ -55,6 +55,11 @@
         <artifactId>spring-cloud-starter-netflix-eureka-client</artifactId>
         <version>${spring.cloud.version}</version>
       </dependency>
+      <dependency>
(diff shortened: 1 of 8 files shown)


alpha/alpha-server/pom.xml
```diff
@@ -79,6 +84,28 @@
       <groupId>org.apache.servicecomb.pack</groupId>
       <artifactId>alpha-spring-cloud-starter-eureka</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.servicecomb.pack</groupId>
+      <artifactId>alpha-spring-cloud-starter-consul</artifactId>
```
WillemJiang: Not sure if the eureka and consul starter can live together.

coolbeevip: The default two are disabled and can be enabled separately by parameters.

PR #432, opened 2019-03-16
[SCB-836] Update to support the nested SagaStart

zhfeng:
https://issues.apache.org/jira/browse/SCB-836

I just raise this PR to include my changes about the nested SagaStart and yes it needs the tests to see if it works. Anyway, I'd like to ask @WillemJiang to review it at first since I'm not very sure that I made the right change on the findFirstByTypeAndSurrogateIdGreaterThan to find all of the transactions which need to be compensated when the parent saga is cancelled. I have to say that these SQL is too complex to understand and it could be better to add the explanation. I think it would be very helpful.

Thanks in advance !

### DIFF
--- a/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/SagaStartAspect.java
+++ b/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/SagaStartAspect.java
@@ -69,6 +69,9 @@ public class SagaStartAspect {
   }
 
   private void initializeOmegaContext() {
+    if (context.globalTxId() != null) {
+      context.setParentTxId(context.globalTxId());
+    }
     context.setLocalTxId(context.newGlobalTxId());
   }
 }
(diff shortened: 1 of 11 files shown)


WillemJiang:
It's a good start, we just need to add some test to verify the changes.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/SagaStartAspect.java
```diff
@@ -69,6 +69,9 @@ Object advise(ProceedingJoinPoint joinPoint, SagaStart sagaStart) throws Throwab
   }
 
   private void initializeOmegaContext() {
+    if (context.globalTxId() != null) {
+      context.setParentTxId(context.globalTxId());
+    }
```
zhfeng: @WillemJiang my changes is here.

WillemJiang: I think we should set the parentTxId from the localTxId, as the globalTxId won't be changed.

zhfeng: No, I don't think so. The nested saga could be cancelled individually that means it should have the different the globalTxId.
```
SagaStart (globalTxId = 1, localTxId = 1, parentTxId = null)
Nested SagaStart (globalTxId = 2, localTxId = 1, parentTxId = 1)
Cancel Nested Saga ( this could rollback the globalTxId = 2)
Close Saga ( this could close the globalTxId = 1

zhfeng: The other case is when parent saga is cancelled, it should cancel the nested saga whenever it close before or not.

zhfeng:
I withdraw this PR and it is not very clear since we have the FSM to handle the events currently. Maybe we need a new event to start a nested saga.

PR #433, opened 2019-03-18
[SCB-1204] acceptance test with consul

coolbeevip:


coolbeevip:
> We need update the License file version at the same time.

Ok, I will remind you after I change it.

PR #452, opened 2019-04-10
[SCB-1247] Set default value in alpha.cluster.address on omega side

coolbeevip:
When no alpha.cluster.address is configured on the omega side, the default value is 0.0.0.0:8080


loveoobaby:
@ConditionalOnProperty(value = {"alpha.cluster.address"})   must be delete! 
but when alpha.cluster.register.type=default, I think alpha.cluster.address must be set, this is the right way to use omega. if it is not set, the omega procession should start failed, the default value is useless.

Issue #459, opened 2019-05-08
pack tcc 调用第2个服务时报错


jinbinguo:
使用java-chassis的consumer 和 provider， TCC事务的参与方抛出如下异常。
```
2019-05-08 09:17:21.324  INFO 676 --- [nio-9101-exec-9] o.a.s.qps.QpsControllerManager           : Create qpsController, configKey = [pkms-test]
2019-05-08 09:17:21.324  INFO 676 --- [nio-9101-exec-9] o.a.s.qps.QpsControllerManager           : Create qpsController, configKey = [pkms-test.pkms-test1-pack]
2019-05-08 09:17:21.324  INFO 676 --- [nio-9101-exec-9] o.a.s.qps.QpsControllerManager           : Create qpsController, configKey = [pkms-test.pkms-test1-pack.testAuditBillEntry]
2019-05-08 09:17:21.692 ERROR 676 --- [pool-2-thread-1] o.a.s.p.o.c.g.c.LoadBalanceSenderAdapter : Retry sending event ParticipationStartedEvent{globalTxId='null', localTxId='8944bbe9-a4bd-40ec-aff7-685fc9e51151', parentTxId='null', confirmMethod='public void com.peaksport.pkms.leanrtest1.service.Test1PackService.auditBill1_confirm(java.lang.String)', cancelMethod='public void com.peaksport.pkms.leanrtest1.service.Test1PackService.auditBill1_cancel(java.lang.String)'} due to failure

java.lang.NullPointerException: null
	at org.apache.servicecomb.pack.contract.grpc.GrpcParticipationStartedEvent$Builder.setGlobalTxId(GrpcParticipationStartedEvent.java:845) ~[pack-contract-grpc-0.4.0.jar:0.4.0]
	at org.apache.servicecomb.pack.omega.connector.grpc.tcc.GrpcTccClientMessageSender.convertTo(GrpcTccClientMessageSender.java:146) ~[omega-connector-grpc-0.4.0.jar:0.4.0]
	at org.apache.servicecomb.pack.omega.connector.grpc.tcc.GrpcTccClientMessageSender.participationStart(GrpcTccClientMessageSender.java:74) ~[omega-connector-grpc-0.4.0.jar:0.4.0]
	at org.apache.servicecomb.pack.omega.connector.grpc.tcc.TccLoadBalanceSender$1.apply(TccLoadBalanceSender.java:49) ~[omega-connector-grpc-0.4.0.jar:0.4.0]
[...]

PR #460, opened 2019-05-08
SCB-1279 add OmegaContext attributes verification on Aspect

coolbeevip:
issues #459 
Improved exception prompt


omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/OmegaContext.java
```diff
@@ -72,4 +72,10 @@ public String toString() {
         ", localTxId=" + localTxId.get() +
         '}';
   }
+
+  public void verify(){
+    if(this.globalTxId == null){
+      throw new RuntimeException("OmegaContext globalTxId is empty");
```
WillemJiang: We could add more information about the error, such as "Please check if you setup the pack transport handler rightly."

PR #461, opened 2019-05-08
[SCB-1281] Allows alpha's restful API to be accessed by CORS

coolbeevip:
Configuration parameter alpha.rest.cors.enabled=true

You can run a script test in the chrome console
```javascript
var xhr = new XMLHttpRequest();
xhr.open('GET','http://0.0.0.0:8090/saga/stats');
xhr.send(null);
xhr.onload = function(e){
	var xhr = e.target;
	console.log(xhr.responseText);
}
```

### DIFF
--- a/alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/mvcconfig/WebConfiguration.java
+++ b/alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/mvcconfig/WebConfiguration.java
@@ -18,6 +18,7 @@
 package org.apache.servicecomb.pack.alpha.server.mvcconfig;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


WillemJiang:
I think we could fix it by hosting the front-end UI with the REST API provider at the same time.

coolbeevip:
> I think we could fix it by hosting the front-end UI with the REST API provider at the same time.

Do you mean using a reverse proxy?

coolbeevip:
issues #462

WillemJiang:
> > I think we could fix it by hosting the front-end UI with the REST API provider at the same time.
> 
> Do you mean using a reverse proxy?

How about we put web module into site resource of alpha server?  In this way we can host the front-end UI through alpha server restful service.

PR #486, opened 2019-06-30
[SCB-1335] Upgrade docker-compose file to 3.0

chanjarster:

---

Upgrade demo's docker-compose files from version 2.1 to 3.0.

Please notice that conditional `depends_on` which is add in 2.1 [\[1\]][1] is removed from 3.0 [\[2\]][2]. So I change the following like below:

```yaml
depends_on:
  database:
    condition: service_healthy
```

to like this:

```yaml
depends_on:
  - database
```

I tested the modified docker-compose file on my machine, that change also works except that there are some retry log like this:

```txt
2019-06-30 02:45:56.059 ERROR 1 --- [ault-executor-2] o.a.s.p.o.c.g.c.ReconnectStreamObserver  : Failed to process grpc coordinate command.
stacks...
2019-06-30 02:45:56.291  INFO 1 --- [pool-2-thread-1] .a.s.p.o.c.g.c.PushBackReconnectRunnable : Retry connecting to alpha at alpha-server.servicecomb.io:8080
2019-06-30 02:45:56.333 ERROR 1 --- [pool-2-thread-1] .a.s.p.o.c.g.c.PushBackReconnectRunnable : Failed to reconnect to alpha at alpha-server.servicecomb.io:8080
stacks...
2019-06-30 02:46:02.359  INFO 1 --- [pool-2-thread-1] .a.s.p.o.c.g.c.PushBackReconnectRunnable : Retry connecting to alpha at alpha-server.servicecomb.io:8080
2019-06-30 02:46:02.527  INFO 1 --- [pool-2-thread-1] .a.s.p.o.c.g.c.PushBackReconnectRunnable : Retry connecting to alpha at alpha-server.servicecomb.io:8080 is successful

```

[1]: https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on
[2]: https://docs.docker.com/compose/compose-file/#depends_on

### DIFF
--- a/docker-build-config/docker-entrypoint.sh
+++ b/docker-build-config/docker-entrypoint.sh
@@ -1,5 +1,20 @@
 #!/bin/sh
 
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 wait_for_services() {
   for svc in ${WAIT_FOR_SERVICES//,/ }
   do
--- a/docker-build-config/docker-entrypoint.sh
+++ b/docker-build-config/docker-entrypoint.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+wait_for_services() {
+  for svc in ${WAIT_FOR_SERVICES//,/ }
+  do
+    host=$(printf "%s\n" "$svc"| cut -d : -f 1)
+    port=$(printf "%s\n" "$svc"| cut -d : -f 2)
+    timeout=$(printf "%s\n" "$svc"| cut -d : -f 3)
+    echo "Waiting for service $host:$port(timeout: $timeout seconds) ready"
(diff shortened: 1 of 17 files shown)


WillemJiang:
@chanjarster 
I just checked the document, it looks like health check doesn't work any more,  we need to use [sh-compatible wait-for](https://docs.docker.com/compose/startup-order/) to verify the service availability before starting the docker image.

I think in this way we could avoid the errors you hit.  
BTW, we could remove the health-check part, once we upgrade the docker compose file to 3.x.

chanjarster:
Thanks for the suggestion. If using wait-for in docker-compose file like this:

```yaml
command: ["./wait-for", "alpha:8080", "--", "java", "-jar", "..."]
```

The `"java", "-jar", "..."` part will be a duplicate of `ENTRYPOINT` line in Dockerfile. So maybe I should write a `docker-entrypoint.sh`?

WillemJiang:
> Thanks for the suggestion. If using wait-for in docker-compose file like this:
> 
> ```yaml
> command: ["./wait-for", "alpha:8080", "--", "java", "-jar", "..."]
> ```
> 
> The `"java", "-jar", "..."` part will be a duplicate of `ENTRYPOINT` line in Dockerfile. So maybe I should write a `docker-entrypoint.sh`?

Yeah, I agree, we could add a shell to wrap the wait-for command, and we need to make sure those shell are placed in the PATH.

chanjarster:
@WillemJiang Please review the new commit.

WillemJiang:
It looks good to me. We just need to wait for the CI build.

WillemJiang:
CI build are passed, @coolbeevip @zhfeng do you have any comments on this PR?
@chanjarster 's last commit put the wait-for as the start entry point of docker images, so we can start the service after the depended service is started.

docker-build-config/docker-entrypoint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
```
WillemJiang: We to add the Apache license header for this file.

chanjarster:
@WillemJiang License header added

PR #495, opened 2019-07-11
[SCB 1321] Using StateMachine for tracing the transaction states

coolbeevip:
Alpha
- [x]  State machine design document
- [x]  State machine prototype
- [x]  State machine prototype unit test
- [x]  Receive saga events using the internal message bus
- [x]  State machine integration test
- [x]  Enable state machine support via parameters 
- [x]  Verify Akka persistent 
- [ ]  Verify Akka cluster reliability
- [ ]  Save the terminated transaction data to the database
- [ ]  Support for in-process nested global transactions
- [ ]  Support for cross-process nested global transactions
- [ ]  Support for query terminated transaction data by RESTful API
- [ ]  Support for query running transaction data by RESTful API
- [ ]  Support for query running transaction data by RESTful API
- [ ]  Support for query suspended global transaction by RESTful API
- [ ]  Support for compensating failed sub-transaction by RESTful API
- [ ]  State machine metrics collection

 Omega Components
- [x]  Enable state machine support via parameters
- [x]  State machine calls omega side compensation
- [x]  @SagaStart supports thread termination after the timeout

Alpha & Omega
- [x]  Acceptance-pack-akka-spring-demo pass
- [ ]  Add sub-transaction timeout exception for akka acceptance test
- [ ]  Add compensation failure for akka acceptance test
- [ ]  Add compensation retry success for akka acceptance test 
- [x]  Alpha single node benchmark performance test
- [ ]  Alpha cluster benchmark performance test

### Tools
- [x]  Alpha Benchmark tools


alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/ActiveMQActorEventChannel.java
```diff
@@ -0,0 +1,47 @@
...
+
+  public ActiveMQActorEventChannel(
+      ActorEventSink actorEventSink) {
+    this.actorEventSink = actorEventSink;
+  }
+
+  @Override
+  public void send(BaseEvent event){
+    try{
+      throw new UnsupportedOperationException();
```
WillemJiang: It's already a runtime exception, we don't need to wrap it again.

PR #497, opened 2019-07-12
SCB-1321 clean up the code of alpha fsm

WillemJiang:

---


alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/sink/SagaActorEventSender.java
```diff
@@ -57,6 +57,7 @@ public void send(BaseEvent event) {
       } else {
         ActorSelection actorSelection = system
             .actorSelection("/user/" + event.getGlobalTxId());
+        //TODO We should leverage the async API that actor provides to send out the message
```
coolbeevip: If we use async API, then we guarantee the order of sub-transactions within a global transaction

WillemJiang: I think it should be fine, even we send it message one by one with block API, we cannot generate the event order.

coolbeevip: After receiving the event, Alpha first puts it into the channel. SagaActorEventSender gets the event from the channel and sends it. These two steps are asynchronous.

PR #499, opened 2019-07-12
[SCB-1376] Use the internal event to trigger SagaActor stop

coolbeevip:


coolbeevip:
> It looks when SagaActor is in the Suspended status then it will switch stopped stated.
> What if we need to keep the status information of Suspended into DB first?

SagaActor will call stopSagaData method of SagaDataExtension after stopped, we can persist here.

> What if there is another event is fired which could bring the Suspended SagaActor into other status.

There are two possibilities for entering the suspended state, the first is a timeout, and the second is an abnormal event sequence.
**timeout to suspended:** Only when the timeout parameter is set will it enter the suspended state because of the timeout. If the timeout parameter is less than the request timeout, an exception will be thrown to the user.
**abnormal event sequence:** We don't need to pay attention to subsequent events

PR #506, opened 2019-07-19
SCB-1385 provide a common way to pass the GID and LID

WillemJiang:


omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/TransactionAspect.java
```diff
@@ -46,6 +48,15 @@ public TransactionAspect(SagaMessageSender sender, OmegaContext context) {
...
   Object advise(ProceedingJoinPoint joinPoint, Compensable compensable) throws Throwable {
     Method method = ((MethodSignature) joinPoint.getSignature()).getMethod();
+    // just check if we need to setup the transaction context information first
+    TransactionContext transactionContext = getTransactionContextFromArgs(joinPoint.getArgs());
+    if (transactionContext != null) {
+      LOG.debug("Updated context {} for globalTxId:{} and localTxId:{}", context,
+          transactionContext.globalTxId(), transactionContext.localTxId());
+      context.setGlobalTxId(transactionContext.globalTxId());
+      context.setLocalTxId(transactionContext.localTxId());
+    }
```
zhfeng: It could be better to restore the origin globalTxId and localTxId if they exists.

WillemJiang: As we just get the transactionContext from business object, we need to pass them into next invocation as the omega transport does. Any thought？

zhfeng: So in this situation, the origin globalTxId and localTxId should be empty here. Can you log some warning message if the globalTxId is not empty ?

WillemJiang: I agree，we need to warning the user about it.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/TransactionAspect.java
```diff
@@ -59,4 +70,15 @@ Object advise(ProceedingJoinPoint joinPoint, Compensable compensable) throws Thr
       LOG.debug("Restored context back to {}", context);
     }
   }
+
+  TransactionContext getTransactionContextFromArgs(Object[] args) {
+    if (args != null) {
+      for (Object arg : args) {
+        if (arg instanceof TransactionContextWrapper) {
+          return ((TransactionContextWrapper) arg).getTransactionContext();
+        }
```
chanjarster: I think we can add one more detection:

```java
else if (arg instanceof TransactionContext) {
  return ((TransactionContextWrapper) arg);
}
```

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/TransactionContext.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.servicecomb.pack.omega.context;
+
+/**
+ *  This class is holding the Transaction related context which could be use in customer code
+ */
+public class TransactionContext {
+  private String globalTxId;
+  private String localTxId;
+
+  public TransactionContext(String globalTxId, String localTxId) {
```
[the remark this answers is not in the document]
WillemJiang: good point, I will update the code shortly.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/TransactionAspect.java
WillemJiang: Do you mean that user could just pass the TransactionContext as a parament to the subtransaction class?  Yeah, we could support this feature at the meantime.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/TransactionAspect.java
```diff
@@ -46,6 +48,20 @@ public TransactionAspect(SagaMessageSender sender, OmegaContext context) {
   @Around("execution(@org.apache.servicecomb.pack.omega.transaction.annotations.Compensable * *(..)) && @annotation(compensable)")
   Object advise(ProceedingJoinPoint joinPoint, Compensable compensable) throws Throwable {
     Method method = ((MethodSignature) joinPoint.getSignature()).getMethod();
+    // just check if we need to setup the transaction context information first
+    TransactionContext transactionContext = getTransactionContextFromArgs(joinPoint.getArgs());
+    if (transactionContext != null) {
+      if (context.globalTxId() != null) {
```
coolbeevip: If you first determine if context.globalTxId() is not empty, is it better?

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/TransactionContextProperties.java
```diff
@@ -17,9 +17,11 @@
...
 /**
- * Once the user business class implement this TransactionContextWrapper, Omega could extract the TransactionContext instance
- * out of the business class, and set up the OmegaContext before calling sub transaction method.
+ * Once the user business class implement this TransactionContextWrapper, Omega could extract the GlobalTransactionId
+ * and LocalTransactionId out of the business class, and set up the OmegaContext before calling sub transaction method.
  */
-public interface TransactionContextWrapper {
-  TransactionContext getTransactionContext();
+public interface TransactionContextProperties {
+  String getGloableTxId();
```
chanjarster: Typo, should be `getGlobalTxId()`

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/TransactionAspect.java
WillemJiang: Current we just give a chance to let custom to override the transaction context setting as he want.  The check of globalTxId is just for warning the user about the side effect, if the user find out the GID is set, he could chose not to pass the parameter into the sub transaction method.

coolbeevip: Ok, this looks good.

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/TransactionContextProperties.java
WillemJiang: sorry，it's my mistake. I just submit a quick fix for these typos.

chanjarster:
@WillemJiang This PR is LGTM

zhfeng:
Thanks @WillemJiang - LGTM !

PR #510, opened 2019-07-23
[SCB-1389] write docs for explicit passing transaction context

chanjarster:

---


docs/user_guide.md
```diff
@@ -111,7 +111,108 @@ Take a transfer money application as an example:
...
+}
+@Compensable
+public void bar(BarCommand cmd, TransactionContext parentTxContext) {
+  ...
+  // TransactionContext childTxContext = omegaContext.getTransactionContext();
+}
+```
+
+Notice that `bar` method got parent transaction context in parameter list, and got child transaction context from `OmegaContext` in method body. So if you want to passing transaction context to another service, you should pass child transaction context.
+
```
zhfeng: The pack does not support the nested transaction right now. see [SCB-836](https://issues.apache.org/jira/browse/SCB-836) and I raised the https://github.com/apache/servicecomb-pack/pull/432 but it did not finish. So I think we could add a note here and reference to the SCB-836.

chanjarster: @zhfeng  I think what I mentioned here is nested local tx not nested saga though

zhfeng: well, it could be better to use the injected or imported transaction context here.

zhfeng: parent transaction context -> injected (imported) transaction context
child transaction context -> local (exported) transaction context

zhfeng: Anyway, if the users want to use the chained service invoking which means service A -> B -> C, it looks like they would like to pass the explicit transaction context from the parameter list when B -> C.

zhfeng:
Thanks @chanjarster - it looks good to me.

docs/user_guide.md
```diff
@@ -111,7 +111,108 @@ Take a transfer money application as an example:
...
+@Compensable
+public void bar() {
+  ...
+}
+```
+
+Here is how Omega does:
+
+1. Service A's `foo` method opens a new global transaction.
+2. `TransactionClientHttpRequestInterceptor` injects transaction context into request headers when `RestTemplate` request Service B.
```
WillemJiang: Can we add the file link for the interceptor?

docs/user_guide.md
```diff
@@ -111,7 +111,108 @@ Take a transfer money application as an example:
...
+3. When Service B receive the request, `TransactionHandlerInterceptor` extract context info from request headers.
+
+Omega supports following implicity transaction context passing:
+
+1. omega-transport-{dubbo,feign,resttemplate,servicecomb}.
+2. Method call in the same thread (based on `OmegaContext` thread local fields).
+3. `java.util.concurrent.Executor{Service}` annotated by `@OmegaContextAware`.
+
+So here comes a problem: what if implicit transaction context passing can't work? For example, Service A invokes Service B via some RPC library and no extension can be made to injecting or extracting transaction context information. In this situation you need explicit transaction context passing. Omega provides two classes to achieve that.
```
WillemJiang: Please specify the pack version we provide here.

docs/user_guide.md
chanjarster: Resolved

docs/user_guide.md
chanjarster: Resolved

PR #511, opened 2019-07-23
SCB-1374 support fsm redis channel

cmonkey:

---


alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/AbstractActorEventChannel.java
```diff
@@ -42,6 +45,7 @@ public void send(BaseEvent event) {
       this.sendTo(event);
       metricsService.metrics().doEventAccepted();
     } catch (Exception ex) {
+      logger.error("send Exception = [{}]", ex);
```
WillemJiang: log error can show the stack trace if the last object is exception,  but I'm not sure if it works for your case.

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/redis/MessagePublisher.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.servicecomb.pack.alpha.fsm.channel.redis;
```
WillemJiang: License header.

pom.xml
```diff
@@ -83,6 +83,7 @@
     <netty.boringssl.version>2.0.7.Final</netty.boringssl.version>
     <netty.version>4.1.24.Final</netty.version>
     <zookeeper.version>3.4.13</zookeeper.version>
+    <fastjson.version>1.2.58</fastjson.version>
```
WillemJiang: If we add new dependency, we need to update the license file [here](https://github.com/apache/servicecomb-pack/blob/master/distribution/src/release/LICENSE)

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/RedisActorEventChannel.java
```diff
@@ -31,13 +33,17 @@
...
-      ActorEventSink actorEventSink, MetricsService metricsService) {
+      ActorEventSink actorEventSink, MetricsService metricsService, RedisMessagePublisher redisMessagePublisher) {
     super(actorEventSink, metricsService);
+    this.redisMessagePublisher = redisMessagePublisher;
   }
 
   @Override
   public void sendTo(BaseEvent event){
-    throw new UnsupportedOperationException("Doesn't implement yet!");
+    LOG.info("sendTo message = [{}]",event);
```
coolbeevip: I suggest using LOG.debug

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/redis/RedisMessagePublisher.java
```diff
@@ -0,0 +1,43 @@
...
+    private ChannelTopic channelTopic;
+
+    public RedisMessagePublisher(RedisTemplate<String, Object> redisTemplate, ChannelTopic channelTopic) {
+        this.redisTemplate = redisTemplate;
+        this.channelTopic = channelTopic;
+    }
+
+    @Override
+    public void publish(Object data) {
+        logger.info("send message [{}] to [{}]", data, channelTopic.getTopic());
```
coolbeevip: I suggest using 
```java
if(LOG.isDebugEnabled()){
  LOG.debug(xxx)
}
```

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/AbstractActorEventChannel.java
coolbeevip: Which reminds me, here should throw an exception

coolbeevip:
When Alpha cluster, only one RedisMessageSubscriber should be allowed at the same time. Otherwise, it will repeat consumption. 

You can use the following method to determine whether the current node is the master node.

```java
@Autowride
NodeStatus nodeStatus;

if(nodeStatus.isMaster()){
  // Is master
}
```

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/redis/RedisConfiguration.java
```diff
@@ -0,0 +1,60 @@
+package org.apache.servicecomb.pack.alpha.fsm.channel.redis;
```
WillemJiang: We need to add License header here.

PR #517, opened 2019-07-28
SCB-1386 Introduce sendingSagaEnd attribute to SagaStart and Compensable annotations

WillemJiang:


omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
```diff
@@ -72,4 +72,11 @@
    * @return the timeout value
    */
   int timeout() default 0;
+
+  /**
+   * Sending out SagaEnd event to Alpha once the SagaStart annotated method is finished without any error.
+   * Default value is false, which means Omega never send out the SagaEnd event to Alpha once the annotated method is finished.
+   * value is true, means this method is last compensable method need to be called, Omega will send SagaEnd event to Alpha once the method is finished.
+   */
+  boolean sendingSagaEnd() default false;
```
chanjarster: javadoc is related to @SagaStart but this class is @Compensable.
Besides, that looks weird to have a method named sendingSagaEnd here.
Would you like to explain the purpose of this method?

WillemJiang: It just give a chance for the user to send the SagaEnd event out when the local transaction is finished.

chanjarster: As to discussion in SCB-1386, it's hard to decide when to send SagaEnd event on the @Compensable method side.

I think it's enough to keep this feature in @SagaStart, no @Compensable

See [comment-1][1], [comment-2][2]

[1]: https://issues.apache.org/jira/browse/SCB-1386?focusedCommentId=16889501&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16889501
[2]: https://issues.apache.org/jira/browse/SCB-1386?focusedCommentId=16889505&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16889505

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/annotations/SagaStart.java
```diff
@@ -37,4 +37,11 @@
    * @return
    */
   int timeout() default 0;
+
+  /**
+   * Sending out SagaEnd event to Alpha once the SagaStart annotated method is finished without any error.
+   * Default value is true, which means Omega sends out the SagaEnd event to Alpha once the annotated method is finished.
+   * Value is false, which means Omega never sends out the SagaEnd event to Alpha once the annotated method is finished.
+   */
+  boolean sendingSagaEnd() default true;
```
zhfeng: I prefer to use the "delay" attribute name. It could be clear that the saga transaction will be closed later.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
zhfeng: I agree and it could be helpful to introduce the OmegaClient which could be used by the user to control the saga.

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/annotations/SagaStart.java
chanjarster: I prefer `autoClose` or `async`

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
WillemJiang: Current solution just work last transaction is async invocation, if there are more than one async invocations , it's hard to tell when all the local transaction is finished. 
Maybe we can introduce a timeout on Alpha side to verify the finished transactions number, if the all the transactions are committed,  Alpha can mark the Saga transaction is finished.

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/annotations/SagaStart.java
WillemJiang: +1 for autoClose.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/AbstractRecoveryPolicy.java
```diff
@@ -32,11 +32,20 @@ public abstract Object applyTo(ProceedingJoinPoint joinPoint, Compensable compen
...
-      return wrapper.applyTo(joinPoint, compensable, interceptor, context, parentTxId, retries);
+      result = wrapper.applyTo(joinPoint, compensable, interceptor, context, parentTxId, retries);
     }else{
-      return this.applyTo(joinPoint, compensable, interceptor, context, parentTxId, retries);
+      result = this.applyTo(joinPoint, compensable, interceptor, context, parentTxId, retries);
     }
+    if (compensable.sendingSagaEnd()) {
+      // Just send out the SagaEnd event
+      // TODO we may also invoke the callback here to release some resources
+      interceptor.sendSagaEndEvent();
```
coolbeevip: If there are multiple services that need to be compensated, this will cause Alpha to receive duplicate SagaEndedEvents.

coolbeevip: Alpha will receive the following events?

SagaStartedEvent
TxStartedEvent-1
TxEndedEvent-1
TxStartedEvent-2
TxEndedEvent-2
TxStartedEvent-3
TxAbortedEvent-3
SagaAbortedEvent
TxCompensabledEvent-2
SagaEndedEvent
TxCompensabledEvent-1
SagaEndedEvent

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/annotations/SagaStart.java
coolbeevip: +1 for autoClose.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
chanjarster: Since Alpha can't know how many transactions a Saga will have in advance, I think it's hard to select a sensible timeout value.

coolbeevip: The timeout always exists. If you can't evaluate the specific value, you can set the timeout to be smaller than the request timeout. Otherwise, we can dynamically calculate the average request time of the last N and then take the upper limit based on the dynamic distribution ：）

omega/omega-context/src/main/java/org/apache/servicecomb/pack/omega/context/annotations/SagaStart.java
zhfeng: It looks like the "autoClose" wins :)

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
zhfeng: Well, the timeout is not the good solution here and it could make the things much more complicated. So at the first, we should make a clear documentation to clarify that we only support the last transaction is async invocation currently. We could have more discuss on the mailing list to get to a better solution for the async invoking. I think perhaps we could introduce the omega client and let the user  call omega.close() explicitly. I know these low client apis should be implemented carefully and avoiding for the abusing.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/AbstractRecoveryPolicy.java
WillemJiang: SagaEnd just close the Saga transaction, can we just keep it open and let Alpha close after the timeout?
If there is a async invocation, we can send the async transaction start event first to Alpha,  then we can let the Alpha track the whole states to tell if the Saga transaction is OK to close.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
chanjarster: +1 for @zhfeng

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/AbstractRecoveryPolicy.java
zhfeng: But the problem is how the alpha know the transaction is OK to close ? In the async situation, the alpha has no knowledge with the number of the participates, since the TxStartedEvent-n might be always on the road. The timeout is not a good idea as the saga is typically long running transaction.

coolbeevip: > SagaEnd just close the Saga transaction, can we just keep it open and let Alpha close after the timeout?

I don't recommend it, this introduces new uncertainty

> If there is an async invocation, we can send the async transaction start event first to Alpha, then we can let the Alpha track the whole states to tell if the Saga transaction is OK to close.

We define an attribute in the SagaStart annotation to set the number of sub-transactions for this global transaction. So we don't have to care about async problems, Alpha can use internal counters to judge that all sub-transactions have ended.

chanjarster: Don't like sub-transaction number, that makes user side arch hard to change and cause bugs.
BTW, I think we can discuss this issue in the mail list.

coolbeevip:
Maybe later we need to explain the @SagaEnd usage scenario in the documentation.

WillemJiang:
> Maybe later we need to explain the @SagaEnd usage scenario in the documentation.

I will add some javadoc on this annotation.

PR #524, opened 2019-08-01
[SCB-1399] Drop the support of Spring Boot 1.x for ServiceComb-Pack

coolbeevip:


WillemJiang:
Oh, we need to keep branch for Spring Boot 1.x.

PR #531, opened 2019-08-09
[SCB 1411] Add alpha-ui module to provide a simple Web GUI

coolbeevip:


alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/SuspendedType.java
```diff
@@ -0,0 +1,5 @@
+package org.apache.servicecomb.pack.alpha.core.fsm;
```
WillemJiang: License header.

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/repository/model/GlobalTransaction.java
```diff
@@ -36,11 +36,12 @@
   private Date beginTime;
   @JsonFormat(shape = JsonFormat.Shape.NUMBER)
   private Date endTime;
-  private SagaActorState state;
+  private String state;
```
WillemJiang: Is there any reason that we change to String?

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/SagaData.java
```diff
@@ -34,6 +35,8 @@
   private Date endTime;
   private String globalTxId;
   private Date expirationTime;
+  private int timeout;
+  private SuspendedType suspendedType = SuspendedType.NULL;
```
WillemJiang: What's the meaning of NULL.

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/api/APIv1Controller.java
```diff
@@ -31,33 +32,36 @@
...
+  }
+
+  @GetMapping(value = "/transaction/statistics")
+  ResponseEntity<Map<String,Long>> getTransactionStatistics() {
+    return ResponseEntity.ok(APIv1Impl.getTransactionStatistics());
+  }
+
+  @GetMapping(value = "/transaction/slow")
+  ResponseEntity<List<GlobalTransaction>> getSlowTransactions() {
+    return ResponseEntity.ok(APIv1Impl.getSlowTransactions());
```
WillemJiang: Do you need to add the number to the slow transactions?

alpha/alpha-ui/src/main/resources/static/scss/_buttons.scss
```diff
@@ -0,0 +1,52 @@
+.btn-circle {
```
WillemJiang: License header.

alpha/alpha-ui/src/main/resources/static/scss/_cards.scss
```diff
@@ -0,0 +1,36 @@
+// Custom Card Styling
```
WillemJiang: License header.

alpha/alpha-ui/src/main/resources/static/scss/_charts.scss
```diff
@@ -0,0 +1,29 @@
+// Area Chart
```
WillemJiang: License header.

alpha/alpha-ui/src/main/resources/static/scss/_dropdowns.scss
```diff
@@ -0,0 +1,21 @@
+// Custom Dropdown Styling
```
WillemJiang: License header. If these file is from other place we need to specify the source.

coolbeevip: > License header. If these file is from other place we need to specify the source.

These files are already included in webjars bootstrap-4.3.1.jar, I will remove the scss file from resources/static

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/repository/model/GlobalTransaction.java
coolbeevip: > Is there any reason that we change to String?

Because SagaActorState is defined in alpha-fsm, this causes alpha-fsm and alpha-core cross-references

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/SagaData.java
coolbeevip: > What's the meaning of NULL.

The value of suspendedType is SuspendedType.NULL when the transaction is not in the SUSPENDED state, Or SuspendedType.NONE is clearer?

WillemJiang: Yeah， NONE is much better.

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/api/APIv1Controller.java
coolbeevip: > Do you need to add the number to the slow transactions?

Yes, I better add a size parameter default value is 10

PR #540, opened 2019-08-16
SCB-1442 Topic initialization & Class KafkaMessagePublisher needs to use globalTxId as the partitioning strategy

cmonkey:

---

### DIFF
--- a/alpha/alpha-fsm-channel-kafka/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/kafka/KafkaMessagePublisher.java
+++ b/alpha/alpha-fsm-channel-kafka/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/kafka/KafkaMessagePublisher.java
@@ -18,6 +18,7 @@
 package org.apache.servicecomb.pack.alpha.fsm.channel.kafka;
 
 import org.apache.servicecomb.pack.alpha.core.fsm.channel.MessagePublisher;
+import org.apache.servicecomb.pack.alpha.core.fsm.event.base.BaseEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.kafka.core.KafkaTemplate;
@@ -43,7 +44,12 @@ public class KafkaMessagePublisher implements MessagePublisher {
         }
 
         try {
-            kafkaTemplate.send(topic, data).get();
+            if(data instanceof BaseEvent) {
+                BaseEvent event = (BaseEvent) data;
+                kafkaTemplate.send(topic, event.getGlobalTxId(), event).get();
+            }else{
+                kafkaTemplate.send(topic, data).get();
+            }
         } catch (InterruptedException | ExecutionException e) {
             logger.error("publish Exception = [{}]", e.getMessage(), e);
             throw new RuntimeException(e);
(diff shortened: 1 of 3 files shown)


alpha/alpha-fsm-channel-kafka/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/kafka/KafkaMessagePublisher.java
```diff
@@ -43,7 +44,12 @@ public void publish(Object data) {
         }
 
         try {
-            kafkaTemplate.send(topic, data).get();
+            if(data instanceof BaseEvent) {
+                BaseEvent event = (BaseEvent) data;
+                kafkaTemplate.send(topic, event.getGlobalTxId(), event).get();
+            }else{
+                kafkaTemplate.send(topic, data).get();
```
coolbeevip: If the message format does not meet expectations, we only need to print the error log and discard the message.

cmonkey: This is used internally and should not be called externally, just because the parameters of the publish method defined in MessagePublisher define Object, so you need to convert the input object when you need to get the globalTxId.

coolbeevip: > This is used internally and should not be called externally, just because the parameters of the publish method defined in MessagePublisher define Object, so you need to convert the input object when you need to get the globalTxId.

So I understand that kafkaTemplate.send(topic, data).get() will never be executed

cmonkey: fix check msg type

alpha/alpha-fsm-channel-kafka/src/main/java/org/apache/servicecomb/pack/alpha/fsm/channel/kafka/KafkaMessagePublisher.java
```diff
@@ -43,10 +44,15 @@ public void publish(Object data) {
         }
 
         try {
-            kafkaTemplate.send(topic, data).get();
-        } catch (InterruptedException | ExecutionException e) {
+            if(data instanceof BaseEvent) {
+                BaseEvent event = (BaseEvent) data;
+                kafkaTemplate.send(topic, event.getGlobalTxId(), event).get();
```
WillemJiang: Why do we need to call the Future get method?
 I don't think we need to make the async invocation to be a sync invocation.

coolbeevip: > Why do we need to call the Future get method?
> I don't think we need to make the async invocation to be a sync invocation.

We need to make sure the message is delivered, When the transmission fails, we need to throw an exception to the omega side.

WillemJiang: It will introduce some performance issue,  we should revisit it later.

coolbeevip: > It will introduce some performance issue, we should revisit it later.

I think need a trade-off in reliability and performance, I sent a message to discuss this issue 
https://lists.apache.org/thread.html/f4bb944c2d6b3ccd55fe8c0ad1e1d000e94bba09277ed842a1dff480@<dev.servicecomb.apache.org>

https://lists.apache.org/thread.html/f4bb944c2d6b3ccd55fe8c0ad1e1d000e94bba09277ed842a1dff480@%3Cdev.servicecomb.apache.org%3E

PR #564, opened 2019-09-27
[SCB-1368] Scalability using Sharding from Akka Cluster

coolbeevip:
1. manage to distribute actors across several nodes in the cluster
2. SagaActor automatically recovers when JVM crashes


alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/FsmAutoConfiguration.java
```diff
@@ -20,53 +20,50 @@
...
 @ConditionalOnProperty(value = {"alpha.feature.akka.enabled"})
 public class FsmAutoConfiguration {
 
-  @Value("${alpha.feature.akka.channel.memory.size:-1}")
-  int memoryEventChannelMemorySize;
-
-  @Value("${alpha.feature.akka.transaction.repository.elasticsearch.batchSize:1000}")
+  // TODO
+  //  Size of bulk request, When this value is greater than 0, the batch data will be lost when the jvm crashes.
+  //  In the future, we can use Kafka to solve this problem instead of storing it directly in the ES.
```
zhfeng: It could be useful to raise a JIRA issue for this feature and mark it as optional.

alpha/alpha-server/pom.xml
```diff
@@ -201,6 +201,11 @@
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-all</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>2.1.1</version>
```
zhfeng: please add this in the dependency managements and use the property like ${kafka.version}

zhfeng:
Thanks @coolbeevip for this wonderful contribution and I just request some minor changes.

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/FsmAutoConfiguration.java
coolbeevip: > It could be useful to raise a JIRA issue for this feature and mark it as optional.

Ok，I will create a JIRA for this feature

alpha/alpha-server/pom.xml
coolbeevip: Thank you for reminding me，I have just fixed it

zhfeng:
The CI looks like failing with the FSM unit tests.

coolbeevip:
> The CI looks like failing with the FSM unit tests.

I have no problem with local testing. I think it may be that the use of Akka's asynchronous causes the timeout setting to be unreasonable. I will try to modify the timeout period.

WillemJiang:
Just check the CI failure, it is caused by Report submission to Coveralls API failed. 
All  the test are passed.

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/SagaShardRegionActor.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public Object entityMessage(Object message) {
+      return message;
+    }
+
+    @Override
+    public String shardId(Object message) {
+      int numberOfShards = 10; // NOTE: Greater than the number of alpha nodes
```
WillemJiang: This number should be passed as a parameter, do we need to know the alpha instance number at this time?

alpha/alpha-server/src/main/resources/log4j2.xml
```diff
@@ -16,25 +16,26 @@
...
   </Properties>
   <Appenders>
     <Console name="Console" target="SYSTEM_OUT">
       <PatternLayout>
         <Pattern>${LOG_PATTERN}</Pattern>
       </PatternLayout>
     </Console>
-    <RollingFile name="FileAppender" fileName="logs/alpha-server.log"
-      filePattern="logs/alpha-server-%d{yyyy-MM-dd}-%i.log">
+    <RollingFile name="FileAppender" fileName="logs/${sys:log-file-name}.log"
```
WillemJiang: Why did you change the log file name?

alpha/alpha-fsm/pom.xml
```diff
@@ -113,21 +108,47 @@
...
+      <artifactId>akka-persistence-typed_2.12</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-slf4j_2.12</artifactId>
     </dependency>
 
+    <!--
+      jmx over http
+      http://0.0.0.0:8090/actuator/jolokia/read/akka:type=Cluster
```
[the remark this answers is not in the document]
coolbeevip: I just want to check more JVM information at any time during the test. I will disable it first.

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/SagaShardRegionActor.java
coolbeevip: > This number should be passed as a parameter, do we need to know the alpha instance number at this time?

I feel that 10 nodes are enough, maybe later we can modify it to be user-configurable

alpha/alpha-server/src/main/resources/log4j2.xml
coolbeevip: > Why did you change the log file name?

Sometimes we need to start multiple processes in the same location, so we can distinguish by the port number in the log name.

coolbeevip: The log file name is alpha-server-${server.port}.log. looks like  alpha-server-8080.log

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/FsmAutoConfiguration.java
coolbeevip: I created JIRA https://issues.apache.org/jira/browse/SCB-1503 for it

PR #580, opened 2019-10-24
SCB-1542 fix Test build error with Alpha Server

chanjarster:

---


distribution/src/release/LICENSE
```diff
@@ -721,3 +721,9 @@ For details, see https://github.com/webjars
 * SockJS-client (org.webjars:sockjs-client:1.0.2 - http://webjars.org)
 * webjars-locator (org.webjars:webjars-locator:0.36 - http://webjars.org)
 * webjars-locator-core (org.webjars:webjars-locator-core:0.35 - http://webjars.org)
+
+===========================================================================
```
WillemJiang: As we include the script in the source bundle, we need to update the License in the root directory.

docker-build-config/wait-for
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
```
WillemJiang: I think we need to include the License header here. It could be better if we push the upstream to update the license header for it.

distribution/src/release/LICENSE
```diff
@@ -723,7 +723,8 @@ For details, see https://github.com/webjars
 * webjars-locator-core (org.webjars:webjars-locator-core:0.35 - http://webjars.org)
 
 ===========================================================================
-This product bundles from wait-for script which is licensed under the MIT license.
+This product bundles from wait-for script which is licensed under 
+the MIT license.
 For details, see https://github.com/eficode/wait-for
```
WillemJiang: As we don't ship this file in the convince binary release, we don't need to include this line in the LICENSE file.

docker-build-config/wait-for
```diff
@@ -1,5 +1,27 @@
 #!/bin/sh
 
+# The MIT License (MIT)
```
WillemJiang: Just checked the [policy](https://www.apache.org/legal/src-headers.html#3party) of ASF license header of the third party, we don't have right the change the license header if we are not the copyright owner or owner's agent. I think we just need to list the file location in the License file for source release.

distribution/src/release/LICENSE
chanjarster: 意思是只要在 项目根下的LICENSE里面添加就行了吗？distribution/src/release/LICENSE不要动是吗

docker-build-config/wait-for
chanjarster: So I should just revert the change, right? And root/LICENSE is already list this script so it's nothing should be done with root/LICENSE, right?

PR #591, opened 2019-11-12
SCB-1585 Optimize the Netty server thread model to split bossEventLoo…

coolbeevip:
…pGroup and workerEventLoopGroup

### DIFF
--- a/alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/GrpcStartable.java
+++ b/alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/GrpcStartable.java
@@ -20,6 +20,8 @@
 
 package org.apache.servicecomb.pack.alpha.server;
 
+import io.netty.channel.nio.NioEventLoopGroup;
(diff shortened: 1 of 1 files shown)


alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/GrpcStartable.java
```diff
@@ -63,15 +64,21 @@ public GrpcStartable(GrpcServerConfig serverConfig, EventBus eventBus, BindableS
       if(unusedPort.isPresent()){
         if (serverConfig.isSslEnable()){
           serverBuilder = NettyServerBuilder.forAddress(
-                  new InetSocketAddress(serverConfig.getHost(), unusedPort.getAsInt()));
+                  new InetSocketAddress(serverConfig.getHost(), unusedPort.getAsInt()))
```
WillemJiang: I  think it could be better to provide a method here to create ServerBuilder.

coolbeevip: done

PR #609, opened 2019-11-30
[SCB-1627] Omega side CallbackContext send the compensation result event to Alpha

coolbeevip:
- [x] Send TxCompensateAckSucceedEvent for compensation successfully 
- [x] Send TxCompensateAckFailedEvent for compensation exception
- [x] Alpha waits for Omega compensation response

### DIFF
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
@@ -24,6 +24,8 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.Map;
 
+import org.apache.servicecomb.pack.alpha.core.exception.CompensateAskFailedException;
+import org.apache.servicecomb.pack.alpha.core.exception.CompensateConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -54,6 +56,11 @@ public class CompositeOmegaCallback implements OmegaCallback {
 
     try {
       omegaCallback.compensate(event);
+    } catch (CompensateConnectException e) {
+      serviceCallbacks.values().remove(omegaCallback);
+      throw e;
+    } catch (CompensateAskFailedException e) {
+      throw e;
     } catch (Exception e) {
       serviceCallbacks.values().remove(omegaCallback);
       throw e;
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
@@ -24,7 +24,7 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.Map;
 
-import org.apache.servicecomb.pack.alpha.core.exception.CompensateAskFailedException;
+import org.apache.servicecomb.pack.alpha.core.exception.CompensateAckFailedException;
 import org.apache.servicecomb.pack.alpha.core.exception.CompensateConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -59,7 +59,7 @@ public class CompositeOmegaCallback implements OmegaCallback {
     } catch (CompensateConnectException e) {
       serviceCallbacks.values().remove(omegaCallback);
       throw e;
-    } catch (CompensateAskFailedException e) {
+    } catch (CompensateAckFailedException e) {
       throw e;
     } catch (Exception e) {
       serviceCallbacks.values().remove(omegaCallback);
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
@@ -25,7 +25,7 @@ public interface OmegaCallback {
   default void disconnect() {
   }
 
-  default void ask(CompensateAckType type) {
+  default void getAck(CompensateAckType type) {
   }
 
   default boolean isWaiting() {
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
@@ -17,7 +17,7 @@
 
 package org.apache.servicecomb.pack.alpha.core;
 
-import org.apache.servicecomb.pack.alpha.core.fsm.CompensateAskType;
+import org.apache.servicecomb.pack.alpha.core.fsm.CompensateAckType;
 
 public interface OmegaCallback {
   void compensate(TxEvent event);
@@ -25,7 +25,7 @@ public interface OmegaCallback {
   default void disconnect() {
   }
 
-  default void ask(CompensateAskType type) {
+  default void ask(CompensateAckType type) {
   }
 
   default boolean isWaiting() {
--- a/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/SagaData.java
+++ b/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/SagaData.java
@@ -21,8 +21,7 @@ import java.io.Serializable;
 import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.servicecomb.pack.alpha.core.fsm.SuspendedType;
 import org.apache.servicecomb.pack.alpha.fsm.SagaActorState;
@@ -40,7 +39,7 @@ public class SagaData implements Serializable {
   private boolean terminated;
   private SagaActorState lastState;
   private AtomicLong compensationRunningCounter = new AtomicLong();
-  private Map<String,TxEntity> txEntityMap = new ConcurrentHashMap<>();
+  private ConcurrentSkipListMap<String,TxEntity> txEntityMap = new ConcurrentSkipListMap<>();
   private List<BaseEvent> events = new LinkedList<>();
 
   public String getServiceName() {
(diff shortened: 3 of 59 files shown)


alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
```diff
@@ -54,6 +56,11 @@ public void compensate(TxEvent event) {
 
     try {
       omegaCallback.compensate(event);
+    } catch (CompensateConnectException e) {
```
WillemJiang: I think we just need to check the CompensateAskFailedException here.
, as the other exception catch part just moves the callback connection.
 I doubt CompensateAskFailedException is a typo, the right one should be CompensateAckFailedException.

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
```diff
@@ -17,9 +17,18 @@
 
 package org.apache.servicecomb.pack.alpha.core;
 
+import org.apache.servicecomb.pack.alpha.core.fsm.CompensateAskType;
```
WillemJiang: How about change the CompensateAskType to CompensateAckType;

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/CompensateAskWait.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.servicecomb.pack.alpha.server.fsm;
+
+import java.util.concurrent.CountDownLatch;
+import org.apache.servicecomb.pack.alpha.core.fsm.CompensateAskType;
+
+public class CompensateAskWait extends CountDownLatch {
```
WillemJiang: Wait is not a noun, how about we just the CountDownLatch instead ？Such as CompensateAckCountDownLatch.

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/GrpcOmegaCallback.java
```diff
@@ -19,32 +19,71 @@
...
+    try {
+      GrpcCompensateCommand command = GrpcCompensateCommand.newBuilder()
+          .setGlobalTxId(event.globalTxId())
+          .setLocalTxId(event.localTxId())
+          .setParentTxId(event.parentTxId() == null ? "" : event.parentTxId())
+          .setCompensationMethod(event.compensationMethod())
+          .setPayloads(ByteString.copyFrom(event.payloads()))
+          .build();
+      observer.onNext(command);
+      compensateAskWait.await();
```
WillemJiang: Do we have timeout here?
I don't want to  block the compensate() calling thread.

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/CallbackContext.java
```diff
@@ -15,13 +15,14 @@
  * limitations under the License.
  */
 
-package org.apache.servicecomb.pack.omega.context;
```
WillemJiang: We need to add a old class with deprecated annotation in org.apache.servicecomb.pack.omega.context to provide backward compatible support.

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
coolbeevip: Ok, It's a typo

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/CallbackContext.java
coolbeevip: > We need to add a old class with deprecated annotation in org.apache.servicecomb.pack.omega.context to provide backward compatible support.

This is a good idea otherwise I will modify a lot of tests cases（Although it has been modified a lot 😢）

I think this feature only needs to support the SAGA state machine. For this, We need to add a parameter `alpha.feature.akka.transactions = saga, tcc`

Currently, this feature is enabled with the following two parameters:

```
alpha.feature.akka.transactions=saga
alpha.feature.akka.enabled=true
```
What do you think?

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
coolbeevip: done

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/CompensateAskWait.java
coolbeevip: done

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/GrpcOmegaCallback.java
coolbeevip: I plan to add compensation timeout and retry in the third step, I think to add two parameters to the @Compensable

compTimeout： compensation timeouts 
compRetries：compensation retries

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/CompositeOmegaCallback.java
coolbeevip: Yes, I will rename CompensateAskFailedException to CompensateAckFailedException.

Because CallbackContext does not throw an exception, when the connection is closed we need to throw a CompensateConnectException.

In addition, CompensateConnectException and CompensateAckFailedException have different compensation failure strategies.

- CompensateConnectException: Immediately call another Omega instance
- CompensateAckFailedException: Retry until the max compRetries is reached before calling another Omega instance

omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/CallbackContext.java
WillemJiang: +1, we can create another JIRA to "enable the protocol separately" to track it.

coolbeevip: I will add the compensation confirm event only in the state machine mode in this PR, It's better for small step iterations.

coolbeevip: done

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/GrpcOmegaCallback.java
coolbeevip: > I need to redefine the @Compensable attribute:
> 
> Rename retries to forwardRetries
> Rename timeout to forwardTimeout
> Add attribute reverseRetries
> Add attribute reverseTimeout

I have emailed this question to dev@servicecomb.apache.org
https://lists.apache.org/thread.html/a81f042ba5cc711c0df0a1fe7144b5ed6e9f20336534ccf67d77dcf1%40%3Cdev.servicecomb.apache.org%3E

WillemJiang: As this change will break the old code, I think it could be better if you create another PR for this part of change.

coolbeevip: > As this change will break the old code, I think it could be better if you create another PR for this part of change.

OK，I  have created a JIRA https://issues.apache.org/jira/browse/SCB-1646 for it

alpha/alpha-server/src/main/java/org/apache/servicecomb/pack/alpha/server/fsm/GrpcOmegaCallback.java
```diff
@@ -19,32 +19,71 @@
...
+    try {
+      GrpcCompensateCommand command = GrpcCompensateCommand.newBuilder()
+          .setGlobalTxId(event.globalTxId())
+          .setLocalTxId(event.localTxId())
+          .setParentTxId(event.parentTxId() == null ? "" : event.parentTxId())
+          .setCompensationMethod(event.compensationMethod())
+          .setPayloads(ByteString.copyFrom(event.payloads()))
+          .build();
+      observer.onNext(command);
+      compensateAckCountDownLatch.await();
```
WillemJiang: I don't think it's a good practices to start a wait without timeout, as the compensate method block the method
I think we could create a mechanism to check the response in an async way without waiting for the response by blocking the calling thread.

coolbeevip: This is good advice, I think it can be solved in three steps

step 1: The default timeout is 30 seconds in this PR
```
compensateAckCountDownLatch.await(10, TimeUnit.SECONDS);
```
step2: Support user-defined timeout in another PR  https://issues.apache.org/jira/browse/SCB-1646

step3: I agree with async，many compensations cause many waiting threads.
Maybe we need to add a state to the state machine, switch to this state after calling compensation, and follow-up processing after receiving compensation ack. It leads to increased complexity. 

I'm not sure if this is better. After using the compensation ack mode, I think that the compensation of the sub-transactions should be performed one by one in reverse order.

alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/OmegaCallback.java
```diff
@@ -17,9 +17,18 @@
...
 
+import org.apache.servicecomb.pack.alpha.core.fsm.CompensateAckType;
+
 public interface OmegaCallback {
   void compensate(TxEvent event);
 
   default void disconnect() {
   }
+
+  default void ask(CompensateAckType type) {
```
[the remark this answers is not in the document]
coolbeevip: done

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/SagaData.java
```diff
@@ -40,7 +39,7 @@
   private boolean terminated;
   private SagaActorState lastState;
   private AtomicLong compensationRunningCounter = new AtomicLong();
-  private Map<String,TxEntity> txEntityMap = new ConcurrentHashMap<>();
+  private ConcurrentSkipListMap<String,TxEntity> txEntityMap = new ConcurrentSkipListMap<>();
```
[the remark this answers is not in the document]
coolbeevip: Do you mean to define a class to wrap txEntityMap?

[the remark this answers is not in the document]
WillemJiang: Yes, then we can still get the entry by using the normal Map method.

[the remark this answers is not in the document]
coolbeevip: I just create a JIRA https://issues.apache.org/jira/browse/SCB-1658  for Improve encapsulation on txEntityMap of SagaData

PR #616, opened 2019-12-18
SCB-1669 Fixed reverse compensation sort bug in FSM

coolbeevip:
The order of execution of reverse compensation should be reversed from the start time of the sub-transaction

### DIFF
--- a/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntities.java
+++ b/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntities.java
@@ -17,20 +17,27 @@
 
 package org.apache.servicecomb.pack.alpha.fsm.model;
 
-import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.ListIterator;
+import java.util.Map;
 import java.util.function.BiConsumer;
 import org.apache.servicecomb.pack.alpha.core.fsm.TxState;
 
 public class TxEntities {
 
-  private ConcurrentSkipListMap<String, TxEntity> entities = new ConcurrentSkipListMap<>();
+  private LinkedHashMap<String, TxEntity> entities = new LinkedHashMap<>();
 
   public void forEach(BiConsumer<String, TxEntity> action) {
     entities.forEach(action);
   }
 
   public void forEachReverse(BiConsumer<String, TxEntity> action) {
-    entities.descendingMap().forEach(action);
+    ListIterator<Map.Entry<String, TxEntity>> iterator = new ArrayList<>(entities.entrySet()).listIterator(entities.size());
+    while (iterator.hasPrevious()) {
+      Map.Entry<String, TxEntity> entry = iterator.previous();
+      action.accept(entry.getKey(),entry.getValue());
+    }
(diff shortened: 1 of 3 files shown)


alpha/alpha-fsm/src/test/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntitiesTest.java
```diff
@@ -0,0 +1,52 @@
...
+    TxEntity t2 = TxEntity.builder().beginTime(new Date(System.currentTimeMillis()+10)).build();
+    TxEntity t3 = TxEntity.builder().beginTime(new Date(System.currentTimeMillis()+20)).build();
+    List<TxEntity> entities = Lists.newArrayList(t3, t2, t1);
+    // t1,t2,t3
+    TxEntities txEntities = new TxEntities();
+    ListIterator<TxEntity> iterator = entities.listIterator(entities.size());
+    while(iterator.hasPrevious()){
+      entities.forEach(t->txEntities.put(UUID.randomUUID().toString(),iterator.previous()));
+    }
+    // t3,t2,t1
```
WillemJiang: What if the TxEntities have the same beginTime?

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntities.java
```diff
@@ -17,20 +17,31 @@
...
 
 public class TxEntities {
 
-  private ConcurrentSkipListMap<String, TxEntity> entities = new ConcurrentSkipListMap<>();
+  private ConcurrentHashMap<String, TxEntity> entities = new ConcurrentHashMap<>();
 
   public void forEach(BiConsumer<String, TxEntity> action) {
-    entities.forEach(action);
+    entities.entrySet().stream()
+        .sorted((v1, v2) -> v1.getValue().getBeginTime().compareTo(v2.getValue().getBeginTime()))
```
WillemJiang: I think this store action will introduce some performance issue, if there are lot of txEntity need to be visited.

coolbeevip: I found that there is no multi-threaded access to the TxEntities class, I think I can use LinkedHashMap to save sub-transaction objects. By default, LinkedHashMap maintains the insertion order

alpha/alpha-fsm/src/test/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntitiesTest.java
coolbeevip: I will use LinkedHashMap instead of ConcurrentHashMap, So this avoids using "beginTime" for secondary sorting during reverse forEach

coolbeevip: done

alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/model/TxEntities.java
coolbeevip: done

PR #621, opened 2019-12-25
[SCB-1646] Add "forward" prefix to @Compensable property retries and timeout

coolbeevip:
Rename @Compensable property retries and timeout with the "forward" prefix

### DIFF
--- a/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
+++ b/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
@@ -39,15 +39,15 @@ import java.lang.annotation.Target;
 public @interface Compensable {
 
   /**
-   * The retires number of the compensable method.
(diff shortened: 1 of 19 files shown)


omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/annotations/Compensable.java
```diff
@@ -39,15 +39,15 @@
 public @interface Compensable {
 
   /**
-   * The retires number of the compensable method.
```
WillemJiang: We need to update the document for it, it should be list in the upgrade guide.

coolbeevip: OK, I will add an upgrade-guide.md

coolbeevip: done

PR #626, opened 2019-12-28
[SCB-1695] Add attribute mode to @Compensable annotation

coolbeevip:
Add attribute `mode` to the @Compensable annotation, which contains three options:

- forward 
- reverse (default)
- combine (first forward then reverse)

### DIFF
--- a/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/RecoveryPolicyFactory.java
+++ b/omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/RecoveryPolicyFactory.java
@@ -17,17 +17,19 @@
 
 package org.apache.servicecomb.pack.omega.transaction;
 
+import org.apache.servicecomb.pack.omega.transaction.annotations.CompensableMode;
+
 public class RecoveryPolicyFactory {
   private static final RecoveryPolicy DEFAULT_RECOVERY = new DefaultRecovery();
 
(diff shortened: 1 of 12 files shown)


omega/omega-transaction/src/main/java/org/apache/servicecomb/pack/omega/transaction/RecoveryPolicyFactory.java
```diff
@@ -17,17 +17,19 @@
...
-   * If retries == 0, use the default recovery to execute only once.
-   * If retries > 0, it will use the forward recovery and retry the given times at most.
-   * If retries == -1, it will use the forward recovery and retry forever until interrupted.
+   * If mode is reverse, it will use the reverse recovery
+   * If mode is forward, it will use the forward recovery
+   * If mode is combine, it will use the first forward then reverse
    */
-  static RecoveryPolicy getRecoveryPolicy(int retries) {
-    return retries != 0 ? FORWARD_RECOVERY : DEFAULT_RECOVERY;
+  static RecoveryPolicy getRecoveryPolicy(CompensableMode mode) {
```
WillemJiang: I doubt if we already implement the combine mode.

coolbeevip: The combine mode is not implemented, because the TxAbortEvent event is not sent to alpha after the forward compensation, I will be in another PR implemente combine mode

PR #631, opened 2020-01-12
[SCB-1696] Optimize state machine compensation retry strategy using reverseRetries and retryDelayInMilliseconds for FSM

coolbeevip:
1. Add attribute `reverseRetries` and `reverseTimeout` to @Compensable
2. Pass reverse compensation parameters `reverseRetries` `reverseTimeout` and `retryDelayInMilliseconds` in TxStartedEvent
3. Use `TxCompensateAckSucceedEvent` and `TxCompensateAckFailedEvent` instead of `TxCompensatedEvent` in FSM
4. Global transaction is suspended after multiple retries failed to compensate

### DIFF
--- a/pack-contracts/pack-contract-grpc/src/main/proto/GrpcTxEvent.proto
+++ b/pack-contracts/pack-contract-grpc/src/main/proto/GrpcTxEvent.proto
@@ -42,9 +42,12 @@ message GrpcTxEvent {
   bytes payloads = 7;
   string serviceName = 8;
   string instanceId = 9;
-  int32 forwardTimeout = 10;
-  int32 forwardRetries = 11;
-  string retryMethod = 12;
+  int32 timeout = 10;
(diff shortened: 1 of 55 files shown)


pack-contracts/pack-contract-grpc/src/main/proto/GrpcTxEvent.proto
```diff
@@ -42,9 +42,13 @@ message GrpcTxEvent {
   bytes payloads = 7;
   string serviceName = 8;
   string instanceId = 9;
-  int32 forwardTimeout = 10;
-  int32 forwardRetries = 11;
-  string retryMethod = 12;
+  int32 timeout = 10;
```
WillemJiang: We need to mention the change of GrpcTxEvent.proto in the release note,  to notify other language client for updating.

PR #638, opened 2020-03-21
[SCB-1735] Support Alpha call compensation timeout strategy

coolbeevip:
Use property reverseTimeout of @Compensable  to set the timeout of the compensation method

### DIFF
--- a/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/SagaActor.java
+++ b/alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/SagaActor.java
@@ -21,17 +21,21 @@ import akka.actor.PoisonPill;
 import akka.actor.Props;
 import akka.cluster.sharding.ShardRegion;
 import akka.persistence.fsm.AbstractPersistentFSM;
+import java.io.PrintWriter;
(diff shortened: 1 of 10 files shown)


alpha/alpha-fsm/src/main/java/org/apache/servicecomb/pack/alpha/fsm/SagaActor.java
```diff
@@ -55,7 +59,7 @@
 
 public class SagaActor extends
     AbstractPersistentFSM<SagaActorState, SagaData, DomainEvent> {
-
+  protected static final int PAYLOADS_MAX_LENGTH = 10240;
```
WillemJiang: We need to be able to configure this parameter from setting environment variable.

coolbeevip: Agree, I found static variable PAYLOADS_MAX_LENGTH in class TxEvent, I'll solve this problem in another PR

coolbeevip: I created JIRA https://issues.apache.org/jira/browse/SCB-1811 for it

PR #641, opened 2020-03-25
[SCB-1811] Configure PAYLOADS_MAX_LENGTH parameter from os environment

coolbeevip:
Configure PAYLOADS_MAX_LENGTH parameter from os environment variables

### DIFF
--- a/pack-common/src/main/java/org/apache/servicecomb/pack/common/Environment.java
+++ b/pack-common/src/main/java/org/apache/servicecomb/pack/common/Environment.java
@@ -17,7 +17,12 @@
 
 package org.apache.servicecomb.pack.common;
 
+import java.lang.invoke.MethodHandles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class Environment {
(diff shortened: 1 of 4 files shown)


pack-common/src/main/java/org/apache/servicecomb/pack/common/Environment.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public Environment() {
+    if (payloadsMaxLength == 0) {
+      String val = System.getenv("PAYLOADS_MAX_LENGTH");
+      if (val == null || val.trim().length() == 0) {
+        payloadsMaxLength = PAYLOADS_MAX_LENGTH;
+      } else {
+        try {
+          payloadsMaxLength = Integer.parseInt(val);
+        } catch (NumberFormatException ex) {
```
WillemJiang: We need to add a log to specify the error and fallback action.

coolbeevip: done

PR #667, opened 2020-06-22
[SCB-2004] Fix compensation retry failure during business service startup

coolbeevip:
* Use message-driven compensation retry instead of recursive compensation
* In Omega, if CallbackContenxt does not contain callbackMethod, send TxCompensateAckFailedEvent to Alpha instead of throwing NPE

### DIFF
--- a/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/event/internal/ComponsitedCheckEvent.java
+++ b/alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/event/internal/ComponsitedCheckEvent.java
@@ -22,10 +22,10 @@ import org.apache.servicecomb.pack.alpha.core.fsm.event.base.TxEvent;
 
 public class ComponsitedCheckEvent extends TxEvent {
 
-  private TxState preState;
+  private TxState preComponsitedState;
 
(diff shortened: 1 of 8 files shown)


alpha/alpha-core/src/main/java/org/apache/servicecomb/pack/alpha/core/fsm/event/internal/ComponsitedCheckEvent.java
```diff
@@ -17,10 +17,17 @@
...
 package org.apache.servicecomb.pack.alpha.core.fsm.event.internal;
 
+import org.apache.servicecomb.pack.alpha.core.fsm.TxState;
 import org.apache.servicecomb.pack.alpha.core.fsm.event.base.TxEvent;
 
 public class ComponsitedCheckEvent extends TxEvent {
 
+  private TxState preState;
+
+  public TxState getPreState() {
```
WillemJiang, line 27: This states is used store the state of componsited command execution state. 
It's better to update the name to reflect the meaning of this variable.

coolbeevip: I renamed that `preState` to `preComponsitedState`

PR #702, opened 2021-03-03
[SCB-2204] Fix acceptance test for an Interim fix only for CI to pass

coolbeevip:


coolbeevip:
Fix `alpha_records_transaction_events.feature`  for an Interim fix only for CI to pass and later trace https://issues.apache.org/jira/browse/SCB-2204 to compensate for the failure count problem

WillemJiang (approved):

PR #722, opened 2021-07-29
Could org.apache.servicecomb.pack:persistence-jpa:0.7.0-SNAPSHOT drop off redundant dependencies to loose weight?

Celebrate-future:
Hi, I am a user of project **_org.apache.servicecomb.pack:persistence-jpa:0.7.0-SNAPSHOT_**. I found that its pom file introduced **_37_** dependencies. However, among them, **_7_** libraries (**_18%_**) have not been used by your project (the redundant dependencies are listed below). Reduce these useless dependencies can help prevent conflicts between library versions. MeanWhile, it can minimize the total added size to projects. It can also help enable advanced scenarios for this project. 
This PR helps **_org.apache.servicecomb.pack:persistence-jpa:0.7.0-SNAPSHOT_** lose weight :) I have tested the revised configuration in my local environment. It is safe to remove the unused libraries.

Best regards

### DIFF
--- a/persistence/persistence-jpa/pom.xml
+++ b/persistence/persistence-jpa/pom.xml
@@ -38,6 +38,61 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot</artifactId>
+        <version>2.1.6.RELEASE</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-autoconfigure</artifactId>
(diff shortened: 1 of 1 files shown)


persistence/persistence-jpa/pom.xml
```diff
@@ -38,25 +38,94 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot</artifactId>
+        <version>2.1.6.RELEASE</version>
```
WillemJiang, line 44: Please use the dependency manager in pom instead of specify the version directly here.
It will help us manage the third party dependencies.

PR #757, opened 2022-05-05
SCB-2477 Use CI friendly maven version

coolbeevip:
You can change the version via the command line when deploying like this:

```shell
./mvnw deploy -DskipTests -Ppassphrase -Prelease -Drevision=0.7.0
```

The rolling version only needs to modify the revision of the main pom file.

```xml
<properties>
    <revision>0.7.0-SNAPSHOT</revision>
</properties>
```

### DIFF
--- a/pack-dependencies/pom.xml
+++ b/pack-dependencies/pom.xml
@@ -847,6 +847,7 @@
   <build>
     <pluginManagement>
       <plugins>
+        <!-- Install or deploy artifacts with Maven CI Friendly Versions https://maven.apache.org/maven-ci-friendly.html#install-deploy -->
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>flatten-maven-plugin</artifactId>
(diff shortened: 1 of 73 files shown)


pack-dependencies/pom.xml
```diff
@@ -843,4 +843,52 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>flatten-maven-plugin</artifactId>
```
WillemJiang, line 852: Could you add some comment on why we introduce the faltten-maven-plugin?

coolbeevip: > Could you add some comment on why we introduce the faltten-maven-plugin?

I have added a comment 

> Install or deploy artifacts with Maven CI Friendly Versions https://maven.apache.org/maven-ci-friendly.html#install-deploy

WillemJiang (approved):

